Issue #3, opened 2025-04-18
Add More Enemies


packardbell95:
The game currently has three enemies. Maybe @TheMilkMan6669 has some additional ideas for strange creatures to encounter

packardbell95:
Added Badass Flaming Skeleton, Mysterious Scooter, and Wang Rat in https://github.com/packardbell95/tardquest/pull/7

More to come soon!

PR #7, opened 2025-04-19
Add more enemies

TheMilkMan6669:
Added 3 new enemies; Badass Flaming Skeleton, Mysterious Scooter, and Wang Rat. There will be more in the future, so for now, we're keeping issue #3 open.


src/game.htm
```diff
@@ -227,8 +227,7 @@
             inCombat: false,
             levelingUp: false
         };
-
-        const enemies = [
+const enemies = [
```
packardbell95, line 230: It might be nice to reindent this, but we can handle that later

packardbell95 (approved):

src/game.htm
TheMilkMan6669: Somehow missed that, I'll take note

PR #10, opened 2025-04-19
Demonstrating an alternate approach to styling the battle log

packardbell95:
This is just for discussion and doesn't have to be merged, but I wanted to show an alternate approach to handling the battle log styling with CSS instead of JavaScript


packardbell95:
This PR has already served its purpose as a point of discussion. Closing

PR #22, opened 2025-04-26
Level scaling + various others

### LINKED ISSUE #11: Add enemy level scaling
Increase strength of enemies and EXP gain every certain amount of floors to keep the game from getting braindead easy.

TheMilkMan6669:
Enemies now hit harder, get larger HP pools, and give more EXP + BTC as you delve deeper into the depths of the Tardspire (yes that's what I'm calling it now). In addition to this, I added a bunch of colorization to make the battle log more readable, and fixed the missing DEF stat functionality. Now you should be taking less damage as you level the stat up, as intended. I made double sure the new armor honors the defense increases, though I'm not 100% confident that EVERYTHING is working as intended. It seems to be though. Feel free to test different level scaling values and such, I didn't spend too much time finding a good balance.

`            const floorBoost = Math.floor(floor / 2); // Scale every 2 floors
            if (floorBoost > 0) {
                currentEnemy.hp += floorBoost * 5; // +5 HP per scaling
                currentEnemy.attack = [
                    currentEnemy.attack[0] + (floorBoost * 3), // Adjust attack range based on floor scaling
                    currentEnemy.attack[1] + (floorBoost * 3)  // Adjust attack range based on floor scaling
                ];
                currentEnemy.bitcoins += floorBoost; // Increase Bitcoin drop based on floor scaling
            }`

This should be where you can adjust the level scaling values ^

This would close #11

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -260,7 +260,7 @@
         let seenTiles = Array.from({
             length: HEIGHT
         }, () => Array(WIDTH).fill(false));
-
+        let lastFloorBoostNotice = 0;
         const player = {
             x: 1,
             y: 1,
@@ -268,7 +268,7 @@
             dir: 0,
             hp: 20,
             maxHp: 20,
-            defense: 5,
+            defense: 3,
             persuasion: 5,
             exp: 0,
             level: 1,
@@ -590,11 +590,11 @@
             if (player.inCombat) {
                 const dmg = Math.max(20, Math.round(Math.random() * 10) * 5);
                 currentEnemy.hp -= dmg;
-                updateBattleLog(`You exploded the shit out of ${currentEnemy.name} for ${dmg} HP!!!`);
+                updateBattleLog(`You exploded the shit out of <span style="color: #EC4134;">${currentEnemy.name}</span> for <span style="color: #EC4134;">${dmg} HP</span>!!!`);
                 endOfPlayerTurn();
             } else {
                 explodePath();
-                updateBattleLog("KABOOM! The dungeon walls crumble like charred toast!");
+                updateBattleLog('<span style="color: #EC4134;">KABOOM!</span> The dungeon walls crumble like charred toast!');
                 render();
             }
         }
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -590,11 +590,11 @@
             if (player.inCombat) {
                 const dmg = Math.max(20, Math.round(Math.random() * 10) * 5);
                 currentEnemy.hp -= dmg;
-                updateBattleLog(`You exploded the shit out of ${currentEnemy.name} for ${dmg} HP!!!`);
+                updateBattleLog(`You exploded the shit out of <span style="color: #EC4134;">${currentEnemy.name}</span> for <span style="color: #EC4134;">${dmg} HP</span>!!!`);
```
packardbell95, line 593: Just a heads up: if you use class names instead of inline colors, it would be easier to change them later if we ever want to go for color schemes or anything like that

src/game.htm
```diff
@@ -1704,18 +1795,41 @@
...
+                player.defense += 1;
+            } else if (key === 'p') {
+                player.persuasion += 1;
+            } else {
+                return; // Ignore other keys
+            }
+            
+            // Now level up properly when the player manually chooses to level up
+            player.level++;
+            player.exp = 0; // Reset EXP after leveling up
```
packardbell95, line 1813: No experience rollover?

src/game.htm
```diff
@@ -1771,6 +1885,13 @@
         function descend() {
             floor++;
             updateBattleLog(`Descending into floor ${floor}...`);
+        
+            const floorBoost = Math.floor(floor / 2);
+            if (floorBoost > lastFloorBoostNotice) {
+                updateBattleLog(`<span style="color: red;">As you descend deeper into the dungeon, you sense greater danger than before</span>.`);
```
packardbell95, line 1891: I like the game indicating a greater sense of danger to the player. Nice touch

packardbell95 (approved):

packardbell95:
Changes look good overall. I tried a couple of runs and the balance seems better than before. I also really dig the new colors

src/game.htm
TheMilkMan6669: One of the biggest issues I ran into while working on this update was whenever you leveled up, it would always put your current EXP 10 or so points before the goal for the next level instead of rolling over to the appropriate amount. So I had to force it to set the player's EXP to 0 upon leveling up :|

packardbell95: Oh, interesting. We'll have to keep an eye on that. Might be something we could fix with a modulo or something later on

src/game.htm
packardbell95: Addressed in https://github.com/packardbell95/tardquest/pull/23

PR #23, opened 2025-04-27
Changed inline colors to class colors

TheMilkMan6669:
In regards to your comment on the previous commit, I changed the inline colors to classes instead. Idk what I was thinking last night lol, this will definitely make things easier to change later


packardbell95 (approved):
Awesome. Thanks for making this update!

Issue #14, opened 2025-04-19
Add trickster gambler NPC


TheMilkMan6669:
You may encounter this NPC at random who will offer you a random stat boost for a high sum of Bitcoin (currency). However, the chances of actually getting said upgrade is rather slim, and you only get one chance per encounter.

PR #24, opened 2025-04-30
Adding trickster gambler NPC

packardbell95:
This PR introduces the trickster gambler NPC

### How the Gambler Works
The gambler is an NPC who is randomly encountered in the dungeon. He won't even bother appearing if you don't have enough cash to play his game (200 BTC), so you better save up.

When you talk to him, you get a chance to roll a pair of dice. If you roll double sixes, you get a random stat boost:

| Stat | Boost Amount |
| --- | --- |
| HP | Either +5 or +10 Max HP (also adds to the player's current HP) |
| Defense | Either +1 or +2 points |
| Persuasion | Either +1 or +2 points |

Whether you play or not, the gambler will run off into the darkness after an encounter

The gambler can be killed with C4, just like the merchant

### Other Changes in This PR
- Death functions for the merchant and gambler now belong to each of their respective objects instead of as global functions. These are called `die()`
- Helper functions have been added for dialog lines from either the merchant or the gambler which will make styling consistent and easier to update. These are `merchantSay()` and `gamblerSay()` respectively
- Fixed a bug where a health tile could potentially appear in the same place as an NPC which would interrupt the NPC interaction. Checking for NPC interactions earlier resolves this, though the player has to step on the same tile again to get the health pickup (we may have to adjust how health drops are handled in the game later, but this isn't a big deal)
- Fixed a minor bug with the level up message's text color

#### Gambler Play Price
The gambler's play price is defined as `gambler.playPrice`. It's currently set to 200. Change this value if you want the game to be more or less expensive to play

#### Useful Console Commands
Here are some helpful console commands for testing out the changes:
##### 💎 Get Tons of Health, Torches, and Cash
```javascript
player.hp = player.maxHp = player.inventory.contents.torch = player.bitcoins = 99999;
```
##### ♟️ Place the Merchant (even if he was killed)
```javascript
merchant.isAlive = merchant.isActiveOnFloor = true; setMerchant(); render();
```

##### ♟️ Place the Gambler (even if he was killed too)
Remember that the gambler only appears if you have enough cash, so this will set the player's bitcoins to make sure that he appears
```javascript
player.bitcoins = Math.max(player.bitcoins, gambler.playPrice); gambler.isAlive = gambler.isActiveOnFloor = true; setGambler(); render();
```

##### ♟️ Place the Merchant and Gambler Side-By-Side
Useful for blowing them up with a single brick of C4
```javascript
gambler.x = merchant.x - 1; gambler.y = merchant.y; render();
```

##### ♟️ Place the Merchant and Gambler Next to the Player
```javascript
merchant.isAlive = gambler.isAlive = merchant.isActiveOnFloor = gambler.isActiveOnFloor = true; merchant.x = player.x + 1; merchant.y = player.y; gambler.x = merchant.x + 1; gambler.y = merchant.y; render(); 
```

#### 🎲 Just Gamble
This will roll the dice. The gambler does not need to be active and the player does not need any money to do this (though the player's bitcoin count may hit negative values)
```javascript
visitTheGambler(); gamble();
```

This resolves https://github.com/packardbell95/tardquest/issues/14

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -51,7 +51,8 @@
         }
 
         #inventory:not(.hidden),
-        #merchant:not(.hidden) {
+        #merchant:not(.hidden),
+        #gambler:not(.hidden) {
             width: 100%;
             height: 100%;
             white-space: pre-wrap;
@@ -65,19 +66,38 @@
         }
 
         #inventoryList::before,
-        #merchantItemList::before {
-            content: "[INVENTORY]";
+        #merchantItemList::before,
+        #gamblerMenu::before {
             display: block;
             text-align: center;
         }
+
+        #inventoryList::before {
+            content: "[INVENTORY]";
+        };
 
         #merchantItemList::before {
             content: "[The MERCHANT]";
         }
 
+        #gamblerMenu:not(.hidden)::before {
+            content: "[The GAMBLER]";
+        }
+
+        #gamblerMenu {
+            margin-bottom: auto;
+        }
+
         #inventoryItemDescription,
-        #merchantItemDescription {
+        #merchantItemDescription,
+        #gamblerItemDescription {
             min-height: 90px;
+        }
+
+        #gambleAnimation {
+            white-space: pre;
+            overflow: hidden;
+            text-align: initial;
         }
 
         #viewportContainer {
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
You went way above and beyond with that diceroll animation, seriously amazing job

packardbell95:
Thanks @TheMilkMan6669. I appreciate it!

Issue #25, opened 2025-05-01
Fix merchant purchases


TheMilkMan6669:
Purchased items no longer seem to appear in your inventory. Not sure when this started happening, probably from either the last (Gambler) update or the one before that. Or at least, that's the case on my end currently.

TheMilkMan6669:
After further testing, it was bugged even in the initial Merchant update? Somehow never noticed

packardbell95:
Good catch. I managed to miss this in a previous update, but [the fix is available now](https://github.com/packardbell95/tardquest/pull/26)

PR #26, opened 2025-05-01
Fixing merchant purchases

packardbell95:
The merchant would take your money, but items would not show up in the player's inventory

This is because the items were being put into `player.inventory` which is the entire inventory plus functional data rather than `player.inventory.contents` which is where the item list actually lives

This PR solves the problem by referencing the correct location for inventory storage after purchase

### Testing
1. Check your inventory and note its contents
2. Visit the merchant
3. Buy items from him
4. Leave
5. Open your inventory again
6. Observe that the content count matches what was seen previously, plus whatever what purchased from the merchant

Resolves https://github.com/packardbell95/tardquest/issues/25

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -1967,7 +1967,7 @@ __\\_______^/
             }
 
             player.bitcoins -= item.price.buy;
-            player.inventory[itemKey] = (player.inventory?.[itemKey] || 0) + 1;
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Seems to be working perfectly now, thanks!

PR #29, opened 2025-05-04
Randomized battle music (base)

TheMilkMan6669:
Adds functionality to the game that allows a random song to play during battle. Adding new tracks has also been made very simple.
So far I've only made one new track, but I'll make more when I feel like it. Feel free to make and add some yourself too


packardbell95 (approved):
Awesome. I like the new track. Kind of makes me think of Earthbound a bit

PR #31, opened 2025-05-06
Music toggle fix + various

### LINKED ISSUE #30: Music toggle bug
Recommend adding a check to see if music is disabled when entering battle.
This check doesn't seem to be required anywhere else.

### Current
```   
player.inCombat = true;
party.forEach(member => member.healedThisBattle = false);
updateBattleLog(`A wild <span class="enemy">${currentEnemy.name}</span> appears!`);
playMusic('battle');
 }
```
### New
```
player.inCombat = true;
party.forEach(member => member.healedThisBattle = false);
updateBattleLog(`A wild <span class="enemy">${currentEnemy.name}</span> appears!`);
if (musicEnabled) {
    playMusic('battle');
}
}
```

TheMilkMan6669:
General list of fixes and changes:
- Music toggle button works as intended; which in turn, caused a crap ton of other bugs, which were also fixed. Basically had to do some code restructuring.
- Bottom-left control hints box no longer disappears when entering any menus (inventory, merchant, gambler)
- Your HP now updates immediately when consuming a healing item
- Changed some persuasion stat scaling values. I felt like it was too easy to persuade too many enemies earlygame. I would appreciate some feedback on this one
- The reset button was reworked so that instead of simply linking to the game page itself, it runs a script to reload the game. This alleviates just a little bit of the hassle for anyone who wants to fork this game, port it to Electron, etc.

This will close #30

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -288,7 +288,7 @@
 </head>
 <body>
   <div class="menus">
-        <a href="https://xxthemilkman69xx.neocities.org/dungeon/title.html">TITLE</a>  <a href="https://xxthemilkman69xx.neocities.org/dungeon/game.html">RESET</a>  <a href="#" id="musicToggle">Toggle Music</a>
+        <a href="https://xxthemilkman69xx.neocities.org/dungeon/title.html">TITLE</a>  <a href="#" id="resetButton">RESET</a>  <a href="#" id="musicToggle">Toggle Music</a>
 
         <p><strong>KEY</strong></p>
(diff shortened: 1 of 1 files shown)


packardbell95 (approved):
Changes look good. I left some feedback on a couple of items but none of it is critical

Trying out the Persuasion tactics on enemies, it is slightly easier to persuade them earlier on (bumping the initial stats up to a 1-in-5 chance instead of something closer to 1-in-7). I think the change is good to try rolling with for now

Awesome job on tackling all of these bugs too!

Issue #27, opened 2025-05-01
Buying Equipment


packardbell95:
We were previously discussing allowing the purchase of weapons and armor from the merchant [(last comment about it here)](https://github.com/packardbell95/tardquest/pull/21#pullrequestreview-2785499316)

Some questions for @TheMilkMan6669: How should we handle this in game? I see a few options, though there could be others:

#### Option 1: Different Sections for Wares
Rather than listing all of the merchant's items, maybe the player selects what they are trying to buy (or even sell). So instead of seeing the list of items right away, the merchant instead prompts for the category:

```
▶ Items
  Weapons
  Armor
```

Each would show the list of available wares. Pressing escape would back out to the previous section so the player could select something else if they wanted

#### Option 2: Include Equipment in the Item List
I was kind of thinking about this one when working on the merchant previously, but I'm not sure about it:

The merchant carries a random assortment of items. Maybe he could also randomly carry a gear upgrade which would be the next level higher than what the player currently has:

```
You have 989 BTC in your wallet

  DOWSING ROD . . . . . . . . 20 BTC
  BRICK OF C-4. . . . . . . . 300 BTC
▶ (WEAPON) BLUDGEONING MACE . 500 BTC
```

Though this intermix might be kind of weird

#### Option 3: Have Multiple Merchants
Maybe in addition to an items merchant, there could be a separate equipment merchant, or even separate weapon and armor merchants that specialize in respective categories of wares. It could make the dungeon seem a bit more lively in terms of NPCs, but also might make the dungeon feel more like a feral bazaar

### Other Details
Whatever scenario, I think the main interaction will be the same: The player only has the option to upgrade gear. Upon purchase, it gets equipped automatically. We could allow for the sale of old gear explicitly or automatically, if at all, which leads me to a final question:

### Can The Player Sell Stuff?
Do we want to give the player the option to sell stuff generally? This might make more sense if there were random item drops in the game rather than just Bitcoin rewards (the only random drop that happens currently is the health pack which is activated instantly, so it can't be carried around)

TheMilkMan6669:
I think option 1 would be the best. I agree with having too many NPCs potentially making the world feel too bloated in a way. Option 2 also just feels like kind of a visually and mechanically messy way to go about it. 
Ans to prevent keybind bloat, we could also apply option 1's menu to the player's inventory as well rather than having a totally separate equipment menu bound to a different key. 

[...]

PR #32, opened 2025-05-07
Buying Equipment, Menu Updates

packardbell95:
This PR adds the ability to buy weapons and armor from the merchant

To make this easier to implement, a new menu system has been introduced that consolidates the menus for the merchant and gambler interactions as well as the inventory system

### Buying Equipment
Equipment is always only upgradable. This means that any gear that is bought will automatically make any equal or inferior gear inaccessible to the player

|  |
| --- |
| Upgrading weapons and armor prevents all inferior equipment from being purchased |

### Menu Interactions
The new menus offer a simple and consistent interface for interaction

| Key | Action |
| --- | --- |
| ↑/W | Move the cursor up |
| ↓/S | Move the cursor down |
| Escape | Back out of the current menu |
| Enter/E/Space | Select the current option |

### Menu Organization
Menus are now all handled through the `menu` object. This object handles the interactions and holds all of the menus available in the game

Each menu lives in [the `menu.menus` object](https://github.com/packardbell95/tardquest/blob/9dc514c3374a92666d218695a7ab97f6bd571092/src/game.htm#L2136) (not a super great name, but whatever). Here, each entry is keyed with an identifier. For example, [the Inventory menu is called `inventory`](https://github.com/packardbell95/tardquest/blob/9dc514c3374a92666d218695a7ab97f6bd571092/src/game.htm#L2137), [the gambler's main menu is called `gambler`](https://github.com/packardbell95/tardquest/blob/9dc514c3374a92666d218695a7ab97f6bd571092/src/game.htm#L2326), and so on

### Opening and Closing Menus
To open a menu, you simply call `menu.open()` with its identifier. For example, to open the inventory, you would call:
```javascript
menu.open('inventory');
```

Closing a menu works similarly, but does not need an identifier since it will close whatever menu is open:
```javascript
menu.close();
```

#### Stacked Menus
The menu system stacks menus. This means that you can open multiple menus and have a trail of what menus came before it. This list is stored in [`menu.breadcrumbs`](https://github.com/packardbell95/tardquest/blob/9dc514c3374a92666d218695a7ab97f6bd571092/src/game.htm#L2016) which is the stack of menus. It is currently only used for the merchant

When visiting the merchant, the player is presented with the merchant's main menu. This lists a few possible options:
- Items
- Weapons
- Armor
- Leave

Say the player chooses "Weapons". This is handled by opening a new menu:
```javascript
menu.open('merchantWeapons'); 
```
This opens the weapons menu and adds it onto the stack. Here, the player can buy weapon upgrades. When they are finished, the player closes the menu which calls `menu.close()`. This pops the top entry off of the stack, taking the player to the previous menu. In this case, it is the merchant's main menu where they can choose another option

#### 📝 A Note About Menu Navigation
We may end up incorporating menus into other parts of the game. Some of these may not want to make use of the stacking feature and may want to just replace the menu so that the player can't go back to a previous screen. For example, if we incorporate more nuanced NPC dialog, we may not want to allow the player to back out of a menu

Developing this further, we might want to consider following [the History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) as an example since this kind of menu navigation is pretty similar to what's allowed in browsers generally

### Defining a New Menu
If you want to add a new menu, there are a few parameters that can be set

Again, each menu should be added to the `menu` system's `menus` object, keyed with a unique identifier. Let's say we wanted to add one called `foo`. It might look like this:
```javascript
foo: {
    title: "FOO", // Will be displayed as "「 F O O 」" at the top of the menu
    onOpen: () => {
        // Perform some special actions when the menu is opened (not when it is returned to after backtracking from closing another menu)
    },
    onClose: () => {
        // Perform some special actions when the menu is closed
    },
    landingHtml: () => {
        // Returns the HTML that is displayed immediately below the title
        return `<img src="https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcQeP835chGu_jniJSyXHpvjUR1R2y3tFH461g&s" />`;
    },
    getOptions: () => {
        // Returns the list of menu items
        // Each item must have an "id" which is passed to select() when chosen by the player
        //   Note that an id of "_exit" is a special identifier that will close the menu so that Escape doesn't need to be used
        // The "displayText" is the text that is displayed for the option
        // The "description" is the text that appears at the bottom of the menu when the cursor is on this item
        // "trailText" is an optional field that will display right-aligned information. This is used to display quantities in the inventory and prices in the merchant and gambler menus
        // "disabled" is an optional boolean that will make the option unselectable
        // "className" is an optional field that sets the HTML class for the line. This is mostly used to style merchandise entries that are equipment downgrades or items that are too expensive

        return [
            {
                id: "go",
                displayText: "🟢 Go",
                description: "GO GO GO!",
            },
            {
                id: "maybe",
                displayText: "🟡 PANIC",
                description: "Go faster or stop immediately!",
            },
            {
                id: "stop",
                displayText: "🔴 STOP",
                description: "Don't go!",
            },
            {
                id: "_exit",
                displayText: "See ya",
                description: "Close this damn menu",
            },
        ];
    },
    select: (selectedOptionId) => {
        // This is passed the ID of the option that was selected in getOptions()
        switch (selectedOptionId) {
            case "go":
                updateBattleLog('GO!!!!!!');
                break;
            case "maybe":
                updateBattleLog('Uhh?');
                break;
            case "stop":
                updateBattleLog('STOP!!!!');
                break;
        }
    },
},
```

After plugging this in, running `menu.open('foo');` will show the result:
|  |
| --- |
| Opening and interacting with the "Foo" menu |

---

This will close https://github.com/packardbell95/tardquest/issues/27

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -51,54 +51,32 @@
             align-content: center;
         }
 
-        #inventory:not(.hidden),
-        #merchant:not(.hidden),
-        #gambler:not(.hidden) {
+        #menu:not(.hidden),
+        #menuList:not(.hidden) {
             width: 100%;
             height: 100%;
             white-space: pre-wrap;
             display: flex;
             flex-direction: column;
+            padding-bottom: 10px;
         }
 
-        #inventoryList,
-        #merchantItemList {
+        #menuList,
+        #menuSelectionDescription {
             flex-grow: 1;
         }
 
-        #inventoryList::before,
-        #merchantItemList::before,
-        #gamblerMenu::before {
+        #menuList::before,
+        #menuSelectionDescription::before {
             display: block;
             text-align: center;
         }
 
-        #inventoryList::before {
-            content: "[INVENTORY]";
-        };
-
-        #merchantItemList::before {
-            content: "[The MERCHANT]";
-        }
-
-        #gamblerMenu:not(.hidden)::before {
-            content: "[The GAMBLER]";
-        }
-
-        #gamblerMenu {
-            margin-bottom: auto;
-        }
-
-        #inventoryItemDescription,
-        #merchantItemDescription,
-        #gamblerItemDescription {
-            min-height: 90px;
-        }
-
-        #gambleAnimation {
+        #animation:not(.hidden) {
             white-space: pre;
             overflow: hidden;
             text-align: initial;
+            width: 100%;
         }
 
         #viewportContainer {
@@ -303,20 +281,12 @@
   <div id="controls"></div>
   <div id="viewportContainer">
     <div id="game"></div>
-    <div id="inventory" class="hidden">
-        <div id="inventoryList"></div>
-        <div id="inventoryItemDescription"></div>
-        <div class="alignRight">Press Escape to exit the Inventory</div>
-    </div>
-    <div id="merchant" class="hidden">
-        <div id="merchantItemList"></div>
-        <div id="merchantItemDescription"></div>
-        <div class="alignRight">Press Escape to leave the Merchant</div>
-    </div>
-    <div id="gambler" class="hidden">
-        <div id="gambleAnimation" class="hidden"></div>
-        <div id="gamblerMenu"></div>
-        <div id="gamblerEscapeMessage" class="alignRight">Press Escape to leave the Gambler</div>
+    <div id="animation" class="hidden"></div>
+    <div id="menu" class="hidden">
+        <div id="menuLanding"></div>
+        <div id="menuList"></div>
+        <div id="menuSelectionDescription"></div>
+        <div class="alignRight">Press Escape to exit</div>
     </div>
   </div>
   <div id="ui">
@@ -352,19 +322,13 @@
             exp: 0,
             level: 1,
             inCombat: false,
-            isVisitingMerchant: false,
-            isVisitingGambler: false,
             levelingUp: false,
             inventory: {
-                isOpen: false,
-                selectionIndex: 0,
-                contents: {
-                    canOfHamms: 2,
-                    cupOfLean: 2,
-                    torch: 2,
-                    brickOfC4: 2,
-                    dowsingRod: 2,
-                },
+                canOfHamms: 2,
+                cupOfLean: 2,
+                torch: 2,
+                brickOfC4: 2,
+                dowsingRod: 2,
             },
             weapon: 'pointyStick',
             armor: 'graphicTee',
@@ -376,7 +340,6 @@
             isActiveOnFloor: false,
             x: WIDTH - 2,
             y: HEIGHT - 2,
-            selectionIndex: 0,
             items: [],
             die: function() {
                 this.isAlive = false;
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Amazing job as always, seems to work great. Now we just need to be able to access an equipment menu from the player inventory so we can manually equip owned items.
This is great though, I can finally say I'll have the current version of the game hosted on my website now that we can purchase equipment.

TheMilkMan6669:
@packardbell95 On the topic of NPCs and a potential nuanced dialogue system, how would you feel about having a forum post on Agora where the community can submit their own ASCII art and dialogue so they could have their own NPCs pop up at random in the game? I think it'd be a fun idea to populate the dungeon with some of our fellow Agorans.

packardbell95:
> @packardbell95 On the topic of NPCs and a potential nuanced dialogue system, how would you feel about having a forum post on Agora where the community can submit their own ASCII art and dialogue so they could have their own NPCs pop up at random in the game? I think it'd be a fun idea to populate the dungeon with some of our fellow Agorans.

I think that would be a great idea

There are some other things that we could consider as a part of that too (that aren't required so it won't block any suggestions that Agorans would make):

### Attribution
We should have some fun way to credit each user who came up with a monster's design. Like maybe some kind of monster manual / Pokédex entry for each monster that's been encountered that has some stats and facts listed, or an in battle attribution like Mallow's "Thought Peek" move in Super Mario RPG. Either of these could have some blurb crediting the original artist/designer:

|  |  |
| --- | --- |
| **Pokémon Pokédex:** Navigating a book of creatures, including facts and stats | **Super Mario RPG:** Mallow's "Thought Peek" showing an enemy's stats and fun message in battle |

## Equipment Stats
> Now we just need to be able to access an equipment menu from the player inventory so we can manually equip owned items.

Currently there's no benefit to doing this since the available weapons and armor only have better stats incrementally. I think we could have more of an incentive around this if there are different stat boosts and things that they might provide. I have some ideas around this:

### Additional Types of Equipment: Rings?
We currently have weapons and armor. However, some games have other types of equipment too which offer stat boosts or abilities. One that comes to mind is [the Dungeon of Doom for the original Macintosh](https://archive.org/details/DungeonOfDoomRevealed) which had equippable rings that offered mysterious stat boosts (though that game loved to not communicate equipment effects and had a lot of cursed items which discourage players from actually using them):
|  |
| --- |
| **Dungeon of Doom:** Equipping a mysterious Ruby ring which has indeterminate consequences |

### Environmental Abilities
It might be cool to have equipment that has special properties outside of battle. Like maybe we have helmets and headgear with one being a Miner's Helmet that offers a small defense boost, but also extends the visibility around the player when exploring the dungeon

To that point, it might be cool to have different themes to different dungeon levels. A couple of ideas off the top of my head:
- An interior ocean that has bridges across towards the exit, but the player can equip some gear to swim through water to get to the end faster and more safely
- Gnarled roots of [Yggdrasil](https://en.wikipedia.org/wiki/Yggdrasil) that block passages that the player can cut through with enchanted blades
- A garden of poisonous plants that emit noxious fumes that mess with the player's directional movement as long as the player is on a toxic tile, which could be avoided with some kind of gas mask helmet
- Water currents that push the player (and maybe boulders/walls) on every turn, but would not push the player if they're wearing heavy boots

### Battle Stats
We could also incorporate other battle stats that would have different balances too, like speed which would determine who attacks first in a turn. Or a more obvious one: some kind of charm to enhance the persuasion stat when equipped, though it wouldn't offer much defense as a tradeoff (like a tie-dye shirt of enchantment)

TheMilkMan6669:
I was originally thinking about simply adding armor oieces that give other stat boosts, but I like your ideas a lot more. The ribgs would give a lot more flexibility. I like the idea of having different kinds of environments too.

Issue #36, opened 2025-05-07
Stop Window Scrolling when Using Arrow Keys


packardbell95:
When playing the game in a window that can scroll, the arrow keys will also cause the page's position to move. For example, pressing the down arrow in a menu to select the next option will also cause the entire page to scroll downwards

This is mostly noticeable when playing [the embedded version](https://xxthemilkman69xx.neocities.org/dungeon/main)

The solution should just be to [use `preventDefault()` on keypress events](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

TheMilkMan6669:
Been meaning to take a stab at fixing this for a little while now, I'll look into it tonight

Issue #13, opened 2025-04-19
Add treasure chests and chest mimic type enemy


TheMilkMan6669:
Chests may spawn in random spots around a floor and hold Bitcoin (currency), a healing potion, or it could be a Dark Souls mimic chest that'll eat you alive if you're not prepared to fight

PR #37, opened 2025-05-08
Chests, chest mimics, and healing tile rework

TheMilkMan6669:
Treasure chests (marked as "T" tiles) now spawn on each floor in a random quantity of 1-5. Landing on a T tile will display a menu utilizing our new menu system, allowing the player to choose whether to open the chest or not. Leaving the chest behind will allow the player to come back to it later if they're feeling paranoid... because each chest placed during map generation ALSO now has a 20% chance of being a Mimic Chest! Obviously, if a chest is marked as a mimic, opening said chest will throw you into battle with one. You do not want to mess with these guys  early on, but if you do manage to defeat one, you will be given a decently sizeable chunk of change. However, you will be granted no EXP.

mimic

And since I felt like the system was extremely outdated, I reworked the healing tile system. It was incredibly easy to abuse the healing tiles by just moving back and forth constantly, because they would generate as the player moved, and NOT during map generation. Now they spawn in fixed spots, and heal you by 30% of your current max HP stat. Now healing items are significantly more important to carry, while still having something fall back on and hopefully scrape by.

This will resolve https://github.com/packardbell95/tardquest/issues/13

I've also gone ahead and fixed https://github.com/packardbell95/tardquest/issues/36 while I was at it.

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -221,7 +221,12 @@
         .player {
             color: #2aff00;
         }
-
+        .healtile {
+            color: #6cff74;
+        }
+        .treasurechest {
+            color: #f19b32;
+        }
         .wall {
             color: #e9ff00;
         }
@@ -297,7 +302,8 @@
         <p><span class="exit">E</span> Exit: Touch to descend further</p>
         <p><span class="merchant">M</span> Merchant: Sells you crap</p>
         <p><span class="gambler">G</span> Gambler: A shady shyster</p>
-        <p>H Healing Tile: Heals you to max HP</p>
+        <p><span class="healtile">H</span> Healing Tile: Heals you to max HP</p>
+        <p><span class="treasurechest">T</span> Treasure Chest: Contains items</p>
         <p><span class="unexplored">?</span> Undiscovered: Who knows?</p>
       </div>
   <div id="partylist"></div>
@@ -501,6 +507,13 @@
                 hp: 7,
                 attack: [3, 4],
                 bitcoins: 4,
+            },
+            {
+                id: "mimic",
+                name: "MIMIC",
+                hp: 50,
+                attack: [6, 13],
+                bitcoins: 150,
             },
         ];
         const defeatMessages = [
@@ -726,7 +739,7 @@
         function breakMap(x, y) {
             if (coordsInBounds(x, y)) {
                 if (MAP[y][x] === '#') {
-                    MAP[y][x] = '.';
+                    MAP[y][x] = '.'; // Destroy walls
                 }
 
                 const killedMerchant =
@@ -743,6 +756,11 @@
                     gambler.x === x && gambler.y === y;
                 if (killedGambler) {
                     gambler.die();
+                }
+
+                if (MAP[y][x] === 'T') {
+                    MAP[y][x] = '.';
+                    updateBattleLog('<span class="action">Gee willikers, you just destroyed a perfectly good treasure chest! Oh well...</span><br>');
                 }
 
                 seenTiles[y][x] = true;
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -297,7 +302,8 @@
         <p><span class="exit">E</span> Exit: Touch to descend further</p>
         <p><span class="merchant">M</span> Merchant: Sells you crap</p>
         <p><span class="gambler">G</span> Gambler: A shady shyster</p>
-        <p>H Healing Tile: Heals you to max HP</p>
+        <p><span class="healtile">H</span> Healing Tile: Heals you to max HP</p>
+        <p><span class="treasurechest">T</span> Treasure Chest: Contains items</p>
```
packardbell95, line 306: The Undiscovered tile seems to get cut off now that another option has been added:


We could extend the height of the key panel or remove this option altogether. Not a big deal either way

src/game.htm
```diff
@@ -726,7 +739,7 @@
         function breakMap(x, y) {
             if (coordsInBounds(x, y)) {
                 if (MAP[y][x] === '#') {
-                    MAP[y][x] = '.';
+                    MAP[y][x] = '.'; // Destroy walls
```
packardbell95, line 742: Great idea to comment this :+1:

src/game.htm
```diff
@@ -848,30 +875,109 @@
...
+                merchant.isActiveOnFloor = merchant.isAlive && Math.random() < 0.25;
+            }
+
             if (merchant.isActiveOnFloor) {
                 setMerchant();
             }
 
             gambler.isActiveOnFloor = gambler.isAlive && Math.random() < 0.25;
-            if (merchant.isActiveOnFloor) {
+            if (gambler.isActiveOnFloor) {
```
packardbell95, line 897: Great catch on this one

src/game.htm
```diff
@@ -848,30 +875,109 @@
...
+
+            if (tile === 'T' || tile === 'M') {
+                player.currentChest = { x, y };
+                menu.open("chest");
+            }
+        }
+
+        function interactWithHealingTile(x, y) {
+            const healAmount = Math.ceil(player.maxHp * 0.3); // Heal 30% of max HP
+            player.hp = Math.min(player.maxHp, player.hp + healAmount);
```
packardbell95, line 969: We have the `healPlayer()` helper function that handles the max health check already:
```suggestion
            healPlayer(healAmount);
```

Though we should probably put that helper onto the `player` object so we could call `player.heal()` instead

src/game.htm
```diff
@@ -1330,7 +1441,20 @@
...
 
+     ______________
+    /    \\      /  \\
+    |  ((O)    ((O)|     __________
+    |______________|    (________(O) - - - -
+    |    /VVVVV\\   |=====[/\\/]
+    |    \\ | | /   |      [ ]
+    \\______| |_____/
+            \\|
+            `;
```
packardbell95, line 1457: Hell yeah. This is unironically the most intimidating mimic I've ever seen in a dungeon crawler

packardbell95 (approved):
Awesome work on these updates. I love that you can not only nuke mimics with C4, but also persuade them to join your party. The mimic art simply rules as well

The health tiles are also a lot nicer than the previous system. Just trying it out, I noticed that I would strategize more about whether or not I'd go after a health tile since I could see them

src/game.htm
TheMilkMan6669: What's really funny is I even TRIED to see if anything got cut off, and I somehow still missed it lmfao

src/game.htm
TheMilkMan6669: Bro just has a gun, fuck it

src/game.htm
packardbell95: Damn, I missed that too. I mean, at least we're catching these things pretty quickly, so it's not too bad

src/game.htm
packardbell95: It might be the most powerful weapon in the game that the player can't have 😂

PR #43, opened 2025-05-12
Added starting character building

TheMilkMan6669:
Now when you begin the game, you can allocate 20 points to your stats. So now you can start with a more fine-tuned character build of your own. The level up screen also now uses the new menu systems we have in place.

A Speed (SPD) stat has also been added. The amount of time you have to wait before you can move forward one space again is determined by this stat. Each SPD point increases your movement speed by 5%.

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -2732,7 +2732,7 @@ You found a treasure chest! Do you want to open it?
                 statAllocation: {
                     title: "STAT ALLOCATION",
                     landingHtml: () => {
-                        return `Build your character using the <span class="action">${player.remainingPoints}</span> points you have been allootted. Use them wisely...`;
+                        return `Build your character using the <span class="action">${player.remainingPoints}</span> points you have been allotted. Use them wisely...`;
                     },
                     getOptions: () => {
                         return [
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -2652,8 +2670,194 @@
...
+                        player.exp = 0; // Reset EXP
+                        player.levelingUp = false;
+                        menu.close();
+                        render();
+                    },
+                },
+                statAllocation: {
+                    title: "STAT ALLOCATION",
+                    landingHtml: () => {
+                        return `Build your character using the <span class="action">${player.remainingPoints}</span> points you have been allootted. Use them wisely...`;
```
packardbell95, line 2735: This should be "allotted"
```suggestion
                        return `Build your character using the <span class="action">${player.remainingPoints}</span> points you have been allotted. Use them wisely...`;
```

packardbell95 (approved):
Changes generally look good. Nice work :+1:

src/game.htm
TheMilkMan6669: oh god damn it

PR #45, opened 2025-05-16
Luck stat added, level up point allocation added

TheMilkMan6669:
I've added a Luck (LUK) stat that increases your critical hit rate, as well as slightly increasing the amount of BTC you can find in chests. This also means the player can now randomly land critical hits on enemies.

Much like the starting character building screen, you can now allocate 3 points towards your stats when you level up.

I've also made various other under-the-hood changes. The biggest one here is the SPD-based movement delay no longer affects other inputs, so you can now confirm menu selections, rotate the character, etc before the movement delay timer is over. Much less clunky, needless to say.

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -158,7 +158,7 @@
             padding: 5px;
             overflow: hidden; 
             font-size: 18px;
-            gap: 7px;
+            font-weight: bold;
             text-align: left;
             line-height: 1.1;
         }
@@ -171,11 +171,9 @@
             left: 620px;
             padding: 5px;
             overflow: hidden; 
-            font-size: 10px;
+            font-size: 13px;
             font-weight: bold;
-            gap: 7px;
-            text-align: left;
-            line-height: 1.02;
+            line-height: 1.0;
         }
 
         #battleLog,
@@ -322,6 +320,9 @@
         .SPD {
             color: #7ec5ee;
         }
+        .LUK {
+            color: #f5b609;
+        }
         .action {
             color: #EC4134;
         }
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -1682,11 +1692,10 @@
         }
 
         function move(direction) {
-            if (player.inCombat || gameOver || player.inputDisabled) {
+            if (player.inCombat || gameOver || player.movementDisabled) {
```
packardbell95, line 1695: There are still references to `inputDisabled` that are no longer used anywhere that should be removed. The `inputDisabled` value in the `player` object should also be renamed to `movementDisabled` so that a default is set

src/game.htm
```diff
@@ -2767,29 +2737,53 @@
...
                         if (selectedOptionId === "_confirm") {
                             if (player.remainingPoints > 0) {
-                                updateBattleLog("You must allocate all points before starting!");
+                                updateBattleLog("You must allocate all points before continuing!");
                                 return;
                             }
-                            finalizeStats();
+                            if (isLevelUpAllocation) {
+                                isLevelUpAllocation = false;
+                                updateBattleLog("Stats increased! Time to pancake some yummy diapers!");
```
packardbell95, line 2756: Real Pamperchu vibes

packardbell95 (approved):
Luck stats look good. I noticed an increase in gold from chests and in critical hit chances during battle. Nice 👍️

src/game.htm
TheMilkMan6669: I mean hey, the game is literally called TardQuest

PR #46, opened 2025-05-20
Added inventory submenus

### LINKED ISSUE #33: Add equipment menu
Add a nested menu within the player's inventory screen where they can select their body or hand slot, then view a list of their owned weapons and armor to equip manually. We also need to make sure the equipment menu is not accessible during combat, and strictly only during exploration.
Example of the process during gameplay:

- "I" to open inventory while not in combat
- See "Items" and "Equipment" as selectable menus, much like the merchant menu
- Select "Equipment" 
- Goes to nested menu showing "BODY: Graphic Tee - DEF 1" and "HAND: Nunchucks - ATK 6-9" 
- Select "BODY" menu > Redirects to list of owned armor pieces 
- Select "Leather Armor - DEF 6", then immediately goes back to previous menu showing the body and hand slots. Item is now equipped.

TheMilkMan6669:
The player is now able to manually equip obtained weapons and armor, meaning submenus were also added to the inventory screen. Now when you open the inventory, you can select the "Equipment" or "Items" menu. I've pretty much removed the "notAnUpgrade" logic from equipment purchases so you can actually own all the pieces you buy, rather than it working like a linear upgrade system. I kept the .muted id though in case we do still want to use it for something in the future. One minor but annoying issue I ran into was that when you first buy something from the merchant, you lose the armor/item you start the game off with. Could not figure out how to fix that, so my sorry workaround was the create new starting equipment that is weaker, but can still be bought for free if you really wanted to for some reason. Do let me know how this can be fixed though.

I think we should have a good setup now for when we eventually add rings to the game.

I also consolidated the stats panel stylings into their own classes to keep that shit clean

This will close #33

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -433,9 +433,11 @@
                 torch: 2,
                 brickOfC4: 2,
                 dowsingRod: 2,
+                fingerNail: 1, //Start game with this in inventory
+                pectoralMass: 1, //Start game with this in inventory
             },
-            weapon: 'fingerNail',
-            armor: 'pectoralMass',
+            weapon: 'fingerNail', //Start game with this equipped
+            armor: 'pectoralMass', //Start game with this equipped
             movementDisabled: false,
         };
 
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -2434,28 +2485,169 @@
...
+                        }
+                    },
+                },
+
+                equipBody: {
+                    title: "EQUIP ARMOR",
+                    landingHtml: () => {
+                    },
+                    getOptions: () => {
+                        const ownedArmor = Object.keys(armor).filter(a => player.inventory[a] > 0 || player.armor === a);
```
packardbell95, line 2626: Just a thought, and this could be something to handle later since this works right now, but we could also split the `player.inventory` into different categories (`items`, `weapons`, `armor`). This would save from having to filter the types of collectibles whenever they're referenced

src/game.htm
```diff
@@ -431,7 +458,7 @@
             buy: function(type, key) {
                 const merchandise = (type === 'item' ? items : (type === 'weapon' ? weapons : armor))?.[key];
                 if (!merchandise) {
-                    console.error(`${itemKey} doesn't exist!`);
+                    console.error(`${key} doesn't exist!`);
```
packardbell95, line 461: Nice catch on this

src/game.htm
```diff
@@ -446,6 +473,7 @@
                 if (type === 'item') {
                     player.inventory[key] = (player.inventory?.[key] || 0) + 1;
                 } else {
+                    player.inventory[key] = (player.inventory?.[key] || 0) + 1;
                     player[type] = key;
                 }
```
packardbell95, line 478: Rather than duplicating the line, we could just adjust the logic to always add the item/equipment to the inventory, then only equip it if it is anything other than an item:
```suggestion
                player.inventory[key] = (player.inventory?.[key] || 0) + 1;

                if (type !== 'item') {
                    player[type] = key;
                }
```

packardbell95 (approved):
I left some minor feedback, but there wasn't anything that seemed critical. Overall, the changes look good

> One minor but annoying issue I ran into was that when you first buy something from the merchant, you lose the armor/item you start the game off with. Could not figure out how to fix that, so my sorry workaround was the create new starting equipment that is weaker, but can still be bought for free if you really wanted to for some reason. Do let me know how this can be fixed though.

I didn't reproduce this, but I suspect that since the player has two references to a piece of equipment (`player.weapon` or `player.armor`, and again in `player.inventory`), the missing entry in `player.inventory` might have interfered with this. So the starting `player.inventory` should be updated to reflect the current equipment. Again, that's just my untested hunch

src/game.htm
```diff
@@ -2434,28 +2485,169 @@
...
+                                description: "Return to the inventory menu",
+                            }
+                        ];
+                    },
+                    select: (selectedOptionId) => {
+                        if (selectedOptionId === "_exit") {
+                            menu.close();
+                        } else {
+                            menu.open(selectedOptionId);
+                        }
```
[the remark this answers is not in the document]
TheMilkMan6669: I was actually meaning to have it so using a consumable item exits all the menus, and equipping a weapon/armor still backs you out to the previous menu, but I totally forgot lol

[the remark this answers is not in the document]
packardbell95: The latest update feels a lot better now. Thanks for taking care of that

packardbell95 (approved):
The new changes look great:
- Trying to buy already-owned equipment just has the merchant call you dumb instead of just being a greedy bastard who takes your money
- Using items exits all menus
- Equipping gear keeps the menu open in case the player wants to keep modifying equipment
- New gear is not auto-equipped anymore

Nice work :+1:

PR #48, opened 2025-05-23
Changes to LUK effects + Gambler fix

TheMilkMan6669:
Increasing LUK now somewhat increases your chances at landing a 12 in the gambler diceroll minigame by 3% per point. I also noticed in the recent past that landing a 12 (which was also COMICALLY rare btw) would not trigger a win state. Something must have broke in an update sometime down the line, but that's been fixed here. Also nerfed the crit chance rate when leveling LUK from 2% down to 1%.

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -1850,7 +1850,7 @@ __\\_______^/
             let dmg = Math.floor(Math.random() * playerWeapon.damage.randomMultiplier) + playerWeapon.damage.base;
 
             // Crit attacks based off LUK stat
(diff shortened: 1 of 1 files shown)


packardbell95 (commented):
> I also noticed in the recent past that landing a 12 ... would not trigger a win state.
I haven't noticed this and am having trouble seeing where this would have come up. But that aside, the changes all appear to be working. I like the new tweaks and weights. These are some good changes :sunglasses:

PR #50, opened 2025-05-29
Pseudo-3D Renderer Demo

packardbell95:
This PR is just to show a concept of rendering the dungeon with a pseudo-3D view

This works by taking the player's current visibility triangle and gathering all of the walls and objects within view. The scene is then built from furthest to closest, drawing each item from the sides towards the middle of the view

Enemy encounters and the speed cap have been disabled for this demo. Enemy encounters should take place within the rendered scene which does not yet happen

There are also a number of issues with this demo in its current state, including:
- Incorrect perspective drawing for walls far away on either side of the screen
- Some character art bleedthrough (all characters are drawn on the same `#game` div. It would be better to draw each level of depth onto its own layer to curb this)
- Awful art generally
- Lots of repeated art entries (this is due to the positioning lookups which works pretty well for walls, but is incredibly redundant for sprites)
- Leftover debug code for things like drawing lines and filled quads using characters which is currently not used

I've also included some of my visual notes for the perspective drawing in `template.txt` so I don't lose it


packardbell95:
We're going to roll with the ASCII version. A cleaner implementation will be provided in a future pull request

PR #53, opened 2025-06-05
Adding raycasted environments

packardbell95:
This is the largest update we've had in a while and probably will have for some time. This changes a few things:

### Raycast Environments
Environments are now rendered in pseudo-3D, similar to classic dungeon crawlers

The renderer also now works with layers rather than drawing everything on a single div. This makes it easier to control the brightness for different depths within the screen, though some objects can vaguely be seen through walls because of how font rendering works. [Since we already have some issues with monospace font rendering,](https://github.com/packardbell95/tardquest/pull/53#issuecomment-2943003885) this might be fixed with a future update

### Rendering Fix (Hopefully this time 🤞)
Apparently there are issues with fractional scaling with most monospaced fonts. Rather than relying on system fonts, [the Deja Vu Sans Mono font face](https://dejavu-fonts.github.io/) has been included with this PR and is applied to the entire game. This is the monospaced font used by Linux Mint by default

### Map Cells are Objects
Previously, map cells were represented by single characters. Now, everything on the map is represented with an object instead. The merchant and gambler are also now included on the map and don't have their coordinates stored separately within their own objects which will make their lookups a lot simpler

### Misc Other Stuff
- There was a bug where the persuasion box would close when pressing "T" while it was open
- Objects and NPCs now leave craters behind after being blown to smithereens
- The player can now strafe left and right with the Q and E keys
- When starting the game, the player now always faces down a hallway instead of with their face pointed at a solid wall. The hallway also now greets them with a "Welcome to The Tardspire" banner that disappears once the player crosses beneath it
- Enemies are now drawn in the scene when fighting
- Added a custom wall behind the player when the game starts to encourage them to journey forward
- Out of bounds areas are now rendered as an ominous black floor
- Enemy encounters now darken the rendered background to emphasize the enemy
- Fixed the gamble animation which was probably broken during a previous merge


packardbell95:
Yeah, the battle graphics are currently cropped too tightly and have a different background than intended. That'll be updated soon

I haven't seen that ASCII shifting issue before. I've been developing with Firefox and occasionally testing in Edge. Do you have any more details for when you've seen that?

TheMilkMan6669:
From what I can gather, it just happens when the Tardspire banner is visible and when there's world objects (chests, NPCs, etc) visible in the distance. Usually the shifting doesn't happen when you're one cell away from an object, when its ASCII artwork is at its most detailed. 
I've only tested using Brave, but I can test on Supermium and Edge later tonight

TheMilkMan6669:
I'm leaning on it being a font related issue too, that's something we'll have to look at for sure

TheMilkMan6669 (commented):
The glory hole is a funny little addition, I like that.

I also no longer seem to be getting that strange horizontal shifting that I was having before, so that's an improvement, at least for the most part. The chest is still wonky from a very specific distance, but otherwise looks good. This has all been consistent across Brave, Supermium, Edge, and Firefox.


I've also just noticed, and I'm not sure if this was present before, but you can actually vaguely see assets behind walls because of the tiny creases between ASCII characters.


The Tardspire banner also seems to have a couple spots in the bottom row that are shifted to the left, and generally looks somewhat cramped into the environment.


So the overall general outline is, we need to address:
- The chest sprite
- Wall creases
- Tardspire banner

Which, as mentioned previously, may very likely be mostly a font related issue.

TheMilkMan6669:
It's gross but I use Windows 11.

packardbell95:
Oh, that's easy then. Just install Linux and the problem will be solved 😎

packardbell95:
I tried out [a different monospaced font](https://int10h.org/oldschool-pc-fonts/fontlist/font?ibm_bios#-) which should have looked the same, but now I'm seeing the waviness with that:
|  |
| --- |
| The Tardspire banner waving in the wind? Not quite |

I noticed that some of the characters have different sizes. They're either rendered as `7.20001 x 6.75` or `7.19995 x 6.75`. But at least this gives me something to run with

TheMilkMan6669 (commented):
This is looking amazing on my end now, glad we were able to pin the issue down. I'm personally happy with how the in-battle environment color fade turned out as well.

https://github.com/user-attachments/assets/fdac3058-d6a0-48f6-9769-085b253a8866


Only suggestions I'd make are, of course, fix the height on the walls on the sides of the screen as you mentioned previously and address the tiny font size in the box where the controls are displayed. After that I would say we should be good to merge this into master, unless there's more things you want to take care of.

src/game.htm
```diff
@@ -3842,6 +3842,7 @@
 
                 const merchandiseText = getArticle(merchandise.name) + merchandise.name;
                 updateBattleLog(`You just bought ${merchandiseText}`);
+                render();
```
packardbell95, line 3845: Nice. Thanks for this. I think this render was previously being handled by the keypress function

packardbell95:
Hell yeah. I'm glad it's looking good on your machine

I just pushed a change to increase the font size of the controls. For the wall height, I can handle that later. I think the number of changes is pretty good for now. This PR is getting pretty big already

packardbell95:
Awesome. Thanks for all of the helpful feedback on this 🙌

Issue #54, opened 2025-06-07
Fix wall height


TheMilkMan6669:
Fix the height of the walls on the sides in the 3D viewport. Currently the top and bottom aren't touching the top/bottom edges of the viewport when they should be.

![Image](https://github.com/user-attachments/assets/1109723d-50a8-4bd3-96d9-c556149da046)

PR #58, opened 2025-06-07
Extending display height to allow side walls to be taller

packardbell95:
After setting a fixed font in https://github.com/packardbell95/tardquest/pull/53, the side walls were truncated. This PR extends the display to occupy 6 more characters in height and adjusts the side wall art to match. So now the walls no longer appear truncated

|  |  |
| --- | --- |
| **Before:** Side walls appear truncated | **After:** Walls are no longer cut off |

Fixes https://github.com/packardbell95/tardquest/issues/54

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -105,7 +105,6 @@
             position: absolute;
             top: 0;
             left: 0;
-            margin-top: 3em;
         }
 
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Looks much better now thanks

Issue #55, opened 2025-06-07
Fix "You died on floor X" message


TheMilkMan6669:
As of the latest major update which introduces a 3D environment, the message that tells you what floor you died on displays behind the viewport render. I think this should just be moved to the battle log instead.

![Image](https://github.com/user-attachments/assets/b9b2b469-168b-44a4-8880-46144e2dbced)

Issue #51, opened 2025-05-29
Add persuasion attempt limit


TheMilkMan6669:
There should be a limit on how many persuasian attempts can be made per bsttle to prevent having too big of a party too early on. I would suggest limiting to 3 attempts, and having a ring that can be purchased that can increase the limit by 1.

PR #59, opened 2025-06-07
Party size + attempt limit, new ring, PRS stat scaling changes

### LINKED ISSUE #56: Lock Player Input While Gambling
When interacting with the gambler, the player can slam on the Enter key and play the game multiple times. The game will reset the gamble animation and will actually produce outcomes for each play. The interaction typically gets the game operations out of order which makes the menu visible in place of the background since the menu isn't being closed at the right time

|  |
| --- |
| Hitting Enter multiple times during the gamble animation |

We just need to lock the player's input as soon as the gamble option is selected, then unlock it once the animation has completed

TheMilkMan6669:
You can now have a maximum of 3 members in your party, and by default, can only make two attempts at persuasion per battle. I've added a ring (Ring of Amplified Audio) that increases the attempt limit to 3. This won't spawn in chests and can only be purchased from the Merchant.

I've also adjusted the PRS stat scaling so that it caps out at 100. Meaning, for example, if you had 12 PRS, your chances at persuading an enemy would be 12%, and so on. I've increased the base PRS stat to 15 to reflect this change.

I've also fixed a couple of those bugs that were brought up. Player input is now locked while gambling and the "You died on floor X" message has been moved to the battle log.

This will close #56 , #55 , and #51 .

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -4221,7 +4221,7 @@
                 description: "A ring with a tiny, yet deceptively loud megaphone. Miniphone? idk. +2 PRS and an PRS attempt",
                 effects: { persuasion: 2 },
                 price: 130,
-                merchantStockChance: 1.2,
+                merchantStockChance: 0.2,
                 chestChance: 0.00,
             },
         };
--- a/src/game.htm
+++ b/src/game.htm
@@ -231,6 +231,10 @@
             white-space: pre-wrap;
             overflow-y: auto;
             justify-content: space-between;
+            font-size: 12px;
+        }
+        #yourparty {
+            font-size: 20px;
         }
         #stats1 {
             height: 338px;
@@ -3736,7 +3740,7 @@
         const baseStats = {
             maxHp: 20,
             defense: 5,
-            persuasion: 7,
+            persuasion: 15,
             speed: 12,
             luck: 2,
         };
@@ -4126,6 +4130,7 @@
                 merchantStockChance: 0.3,
                 chestChance: 0.04,
             },
+
         // HP Boosters
             ringValentines: {
                 name: "VALENTINES DAY RING",
@@ -4133,7 +4138,7 @@
                 effects: { maxHp: 5 },
                 price: 60,
                 merchantStockChance: 0.5,
-                chestChance: 0.1,
+                chestChance: 0.07,
             },
             ringBloodstream: {
                 name: "BLOODSTREAM NOSERING",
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -5305,23 +5322,34 @@
...
-                const baseChance = 0.0;
-                const prsBonus = getEffectiveStat('persuasion') * 0.03;
-                const totalChance = Math.min(0.9, baseChance + prsBonus);
+                persuasionAttempts += 1;
+                if (persuasionAttempts > maxPersuasionAttempts) {
+                    updateBattleLog(`Your words are no longer being heard... <span class="action">you have run out of persuasion attempts.</span>`);
+                    enemyAttack();
+                    render();
+                    return;
+                }
```
packardbell95, line 5331: It might be better to add the `persuasionAttempts` after the check so that the attempt number doesn't just keep increasing when no attempt has actually been made
```suggestion
                if (persuasionAttempts >= maxPersuasionAttempts) {
                    updateBattleLog(`Your words are no longer being heard... <span class="action">you have run out of persuasion attempts.</span>`);
                    enemyAttack();
                    render();
                    return;
                }
                persuasionAttempts++;
```

src/game.htm
```diff
@@ -5305,23 +5322,34 @@
...
+                        updateBattleLog(`<span class="action">${currentEnemy.name}</span> was <span class="friendly">spared</span> and went back home...`);
+                    }
                     player.leaveCombat();
                     currentEnemy = null;
                     playRandomExplorationMusic();
                 } else {
-                    updateBattleLog(`${currentEnemy.name} really, quite genuinely, does not care...`);
+                    updateBattleLog(`<span class="action">${currentEnemy.name}</span> really, quite genuinely, does not care...`);
                     enemyAttack();
                 }
```
packardbell95, line 5355: It might be good to add a message when the player has exhausted all of their persuasion uses

```suggestion
                    updateBattleLog(`<span class="action">${currentEnemy.name}</span> really, quite genuinely, does not care...`);
                    enemyAttack();
                }
                
                if (persuasionAttempts >= maxPersuasionAttempts) {
                    updateBattleLog(`<span class="action">You have run out of persuasion attempts for this fight.</span>`);
                }

```

packardbell95 (approved):
Changes look good overall. I left a couple of suggestions but they're nothing critical

src/game.htm
TheMilkMan6669: It's there, I think you just missed it

packardbell95: I saw it. I mean, it only appears if the player has already run out of attempts and they make another one. It would be helpful to show it in the log as soon as the number of attempts have been expired so they don't waste a turn to learn that they're out. But if they haven't paid attention to the message and keep trying, then there's no problem in still listing it again each time imho

TheMilkMan6669: Ah, I see. Makes sense. I'll take a look at this in a little bit

Issue #39, opened 2025-05-11
Consolidate Global Functions into Related Objects


packardbell95:
We have a few global functions that could be consolidated into related objects. A few examples:

- [`healPlayer()`](https://github.com/packardbell95/tardquest/blob/bd3d9f2cec512f0d5e22a44ae972819b83e0b6b6/src/game.htm#L654-L656) could be moved to [the `player` object](https://github.com/packardbell95/tardquest/blob/bd3d9f2cec512f0d5e22a44ae972819b83e0b6b6/src/game.htm#L339)
- Item functionality could be consolidated in the related item objects where they're referenced already anyways, such as [the dowsing rod calling `useDowsingRod()`](https://github.com/packardbell95/tardquest/blob/bd3d9f2cec512f0d5e22a44ae972819b83e0b6b6/src/game.htm#L628) or [using C4 calls the global `useC4()`](https://github.com/packardbell95/tardquest/blob/bd3d9f2cec512f0d5e22a44ae972819b83e0b6b6/src/game.htm#L648C28-L648C33) when that functionality could just be placed inline directly, especially since nothing else calls those functions
- Merchant and gambler functions exist mostly in the global scope rather than on the [`merchant`](https://github.com/packardbell95/tardquest/blob/bd3d9f2cec512f0d5e22a44ae972819b83e0b6b6/src/game.htm#L364) and [`gambler`](https://github.com/packardbell95/tardquest/blob/bd3d9f2cec512f0d5e22a44ae972819b83e0b6b6/src/game.htm#L414) objects respectively

We _could_ also start wrapping and scoping things like the map itself to help keep it organized too, though we could save that for another discussion. Just some basic organization would satisfy this issue for now

PR #60, opened 2025-06-08
Consolidating global functions, menu updates, random fixes and cleanup

packardbell95:
This is mostly a cleanup PR to kill an old issue. Here's what's new:

### General Updates
- Removed unused code
- Consolidated functions for existing entities (mostly merchant-, gambler-, and player-specific functions now belong to their respective objects)
- Addressed variable scoping issues in switch statements by wrapping cases in blocks
- Fixed sprite positioning for a far wall and the merchant up close
- Adjusted the gambler outcome message so that the dice roll outcome is displayed before the win/loss result

### Menu Updates
- Fixed a bug where the cursor could sometimes be set on entry numbers beyond the range of available entries (eg: going to the last page and selecting the last item would often make the cursor disappear because it's not pointing at a valid entry)
- Moved the pagination counter to the top of the list of items
- Added left and right arrows to the pagination counter to emphasize how to turn the page
- Hiding the pagination counter if there is only one page of options
- The `title` can now be either a string or a function, currently used for the Stat Allocation / Level Up menu. Functions are expected to return a string that is used for the title
- Updated the Stat Allocation / Level Up menu options so that either Diceroll or stats come first instead of Confirm since stat points need to be spent before Confirm can be selected anyways
- Colorizing the stat menu options so that Confirm is red if there are points left to spend, and that stats will turn red once all points have been spent

Closes https://github.com/packardbell95/tardquest/issues/39

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -153,11 +153,11 @@
         #menuList {
             margin-bottom: auto;
             padding-left: 1em;
+            min-height: 150px;
         }
         #pageNum {
-            text-align: center;
-            font-size: 0.9em;
-            margin-top: 5px;
+            text-align: right;
+            margin-right: 14px;
             color: #ccc;
         }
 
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Changes are looking good to me. The colorized stat allocation menus are an excellent addition.

packardbell95:
Awesome. Thank you!

PR #61, opened 2025-06-09
First person animations, movement screen shake

TheMilkMan6669:
I've added first person animations and some mild screen shake effects when your move. I may add more animations at some point in time, but for now I just have two here.

After allocating your stats during initial startup or when leveling up, your character will eat a delicious, nutritious rodent off the ground. This is now our lore reason for getting a full heal every level up.

When equipping a torch, you now visibly carry a torch on-screen. It will go away once you reach the next floor, since obviously that floor hasn't been fully revealed yet.

https://github.com/user-attachments/assets/fc1bb687-5dd6-48db-b5f7-72f1f4fbd9a4

No rubber toy rats were harmed during filming.


packardbell95 (commented):
Hell yeah. This rules!

The only thing I noticed is that sometimes the rat animation and sound effect play at odd times since they're being handled with `setTimeout()`. I created a follow up PR pointed at this one that converts the animation to Webm which means that we don't have to worry about pairing the video and audio together in JavaScript since they live together in the video file, if you want to check that out:
https://github.com/packardbell95/tardquest/pull/62

PR #63, opened 2025-06-10
Impactful Explosions

packardbell95:
This PR adds an animation for the C4 explosion. Walls also are now replaced with rubble when they are blown up

|  |
| --- |
| BOOM! Here comes the BOOM! |

The closest void tile was also too short so I extended it in this PR as well

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -209,6 +209,48 @@
             z-index: 9;
         }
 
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Looks great, I love the leftover rubble too. Very nice touch

Issue #57, opened 2025-06-07
Apply Ring of Sightliness to Rendered Scene


packardbell95:
The Ring of Sightliness extends visibility on the minimap, but does not extend visibility in the rendered game. Equipping it should also extend the rendered view's visibility depth as well

One way we could do this is to cull the current view and reduce it by two squares and/or by making the furthest away more grayed out

PR #64, opened 2025-06-10
Visibility Updates, Etc

packardbell95:
This PR adds the following changes:
- Visibility updates
  - The dungeon is now generally darker and more mysterious
  - Added variable depth
  - Darkness is displayed if visibility isn't maxed out
  - Torch now extends visibility slightly and adds a glow around the player's vision
  - Ring of Sightliness now maxes out visibility and enhances color saturation
  - Walls directly in front of the player are now one level brighter, making the lighting much more accurate
- Torch animation updates
  - The rising torch animation has been replaced with a CSS animation instead, avoiding the need for a separate GIF
  - The torch will now raise and lower when activated and deactivated
- Moved ring equip functionality out of the menus and into `player.equipRing()`
- Added `player.hasEquippedRing()` helper function so we don't need to check both hands each time we're looking for an equipped ring
- The merchant will now have all items in stock when the `/debug` cheat is used
- Consolidated `merchant.setRingWares()` into `merchant.setWares()`
- The game will now stop the player from lighting another torch if they already have one lit
- Consolidated `updateSeenTiles()` functionality to use the existing `revealMapSpot()` function

|  |
| --- |
| Venturing into the darker dungeon |

Closes https://github.com/packardbell95/tardquest/issues/57


TheMilkMan6669 (approved):
Changes all look good, it definitely feels more dungeon-y now. Much more immersive having visibility related items actually affect things visually and not just on the minimap. The new torch raise/lower animation also looks quite nice.

PR #67, opened 2025-06-15
STR+END stats and carry weight system added

TheMilkMan6669:
STR - Each point increases ATK by 1
END - Increases your max carry load
LOAD - Your carry load, duh

Carrying too many items will debuff your SPD. Debuffing starts with 5% when you've reached 50% of your maximum LOAD, and gradually gets worse as you pick up more items, capping out at a 100% debuff. In the future I may very likely make it so weapons and armor take up your LOAD as well, in which case I would need to add a sell menu for the Merchant.

To make the new stats (especially END) less redundant, I've also made it so weapons require a certain amount of STR to equip, same deal with armor and END.

The amount of points given to you during initialization and level ups have been increased. When leveling up, you can now only allocate 2 points per stat at a time. This ensures you're at least somewhat locked into whatever starting build you chose or putting too many points into just one stat.

I've also added a new weapon (DANCE CLUB), added new cheats (/setstr + /setend), and done some rebalancing to the enemy stats, weapon+armor prices, EXP gain, etc.

@packardbell95 If you have any balancing adjustments you think would make more sense, please do contribute. I'm no good at balancing lol

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -7346,7 +7346,7 @@
             // Scale enemy stats based on floor
             const floorBoost = Math.floor(floor / 2); // Scale every 2 floors
             if (floorBoost > 0) {
-                currentEnemy.hp += floorBoost * 7; // +5 HP per scaling
+                currentEnemy.hp += floorBoost * 7; // +7 HP per scaling
                 currentEnemy.attack = [
                     currentEnemy.attack[0] + (floorBoost * 3), // Adjust attack range based on floor scaling
                     currentEnemy.attack[1] + (floorBoost * 3)  // Adjust attack range based on floor scaling
@@ -7394,7 +7394,7 @@
                 } else {
                     const baseExp = 3;
                     const floorBoost = Math.floor(floor / 2);
-                    const exp = baseExp + (floorBoost * 3); // +5 EXP per boost
+                    const exp = baseExp + (floorBoost * 3); // +3 EXP per boost
                     const bitcoinsEarned = currentEnemy.bitcoins;
 
                     const randomMsg = defeatMessages[Math.floor(Math.random() * defeatMessages.length)];
@@ -8557,10 +8557,6 @@
                         return options;
                     },
                     select: (selectedOptionId) => {
-                        if (selectedOptionId === "_back") {
-                            menu.close();
-                            return;
-                        }
                         const weapon = weapons[selectedOptionId];
                         if (!weapon) return;
                         const req = weapon.requiredStr || 0;
@@ -8601,10 +8597,6 @@
                         return options;
                     },
                     select: (selectedOptionId) => {
-                        if (selectedOptionId === "_back") {
-                            menu.close();
-                            return;
-                        }
                         const armorPiece = armor[selectedOptionId];
                         if (!armorPiece) return;
                         const req = armorPiece.requiredEnd || 0;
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -8944,39 +9110,86 @@
...
+                                let statKey = null;
+                                switch (selectedOptionId) {
+                                    case "hp": statKey = "maxHp"; break;
+                                    case "defense": statKey = "defense"; break;
+                                    case "strength": statKey = "strength"; break;
+                                    case "persuasion": statKey = "persuasion"; break;
+                                    case "endurance": statKey = "endurance"; break;
+                                    case "speed": statKey = "speed"; break;
+                                    case "luck": statKey = "luck"; break;
+                                }
```
packardbell95, line 9150: Kind of a nitpick, but since all but one of these are the same, this probably could just be:
```suggestion
                                const statKey = selectionOptionId === "hp" ? "maxHp" : selectedOptionId;
```

src/game.htm
```diff
@@ -8944,39 +9110,86 @@
...
+                                        break;
+                                    case "endurance":
+                                        player.endurance += 1;
+                                        break;
+                                    case "speed":
+                                        player.speed += 1;
+                                        break;
+                                    case "luck":
+                                        player.luck += 1;
+                                        break;
```
packardbell95, line 9187: These could technically be consolidated since they're being referenced by keys that match the `selectedOptionId` and are increasing it by the same amount:
```suggestion
                                    case "defense":
                                    case "strength":
                                    case "persuasion":
                                    case "endurance":
                                    case "speed":
                                    case "luck":
                                        player[selectedOptionId]++;
                                        break;
```

src/game.htm
```diff
@@ -7304,9 +7392,9 @@
                     updateBattleLog(`The <span class='enemy'>MIMIC</span> has been defeated! You find <span class="BTC">${currentEnemy.bitcoins} BTC</span> in its remains!`);
                     player.bitcoins += currentEnemy.bitcoins; // Award scaled BTC
                 } else {
-                    const baseExp = 5;
+                    const baseExp = 3;
                     const floorBoost = Math.floor(floor / 2);
-                    const exp = baseExp + (floorBoost * 5); // +5 EXP per boost
+                    const exp = baseExp + (floorBoost * 3); // +5 EXP per boost
```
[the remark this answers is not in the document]
packardbell95: Addressed in https://github.com/packardbell95/tardquest/pull/67/commits/565cb0e2983c2369d963824095992fe18fbe5127

packardbell95:
Changes look great overall

I like the load mechanic. It really makes the player feel overburdened with stuff when they're carrying too much. I'm also on board with making equipment contribute to the weight, but only after we implement selling as mentioned in the description of this PR. I also wonder if we should add a couple of other things around this mechanic too. Some initial ideas:
- Definitely highlight the numerator portion of the LOAD stat in the UI with red once the weight limit has been met or exceeded
- Maybe we can adjust the breakable floor tiles so that they take three or more steps before they break, but reduce this number with the more weight that the player is carrying (eg: if the player is carrying 50k weight or more, drop the number to 2 max steps)
- Should we start playing labored fat guy breathing noises with each step during movement when the load has been met/exceeded too? Could be funny, plus a way to further indicate that the load mechanic is taking a toll on the player

Regarding the stat balance, the current setup with this PR feels pretty good. I like that only two points can be attributed to a given stat per level up. It makes it harder to just dump everything into a single stat. Incorporating minimum stat requirements for equipment is also a great way to influence players to buff up other stats. For my own runs on this PR, I've noticed that I go for STR + END + LUK early on, but start putting more points towards other stats as I get deeper into the dungeon and need to upgrade my gear or start watching my HP more closely

The Dance Club weapon is also good and nicely bridges the BTC + ATK gap between the Bludgeoning Mace and the Cathode Ray Tube Monitor

I added a commit to this PR to fix a couple of comments and to remove unused menu item entries. The rest of the feedback isn't important otherwise

TheMilkMan6669:
I was having issues with the whole highlight numerator thing, but I'll take another look at that tonight. Breathing sounds are a good idea too, I'll also get on that. The breakable floor adjustments sound like a good change too, maybe we could even make it so if you're  REALLY heavy you'll just fall down on the first step.

PR #69, opened 2025-06-17
Breathing SFX, weight-based cracked floor tiles, QOL, etc

TheMilkMan6669:
Your character starts breathing once you've reached 50% of your max LOAD, and breathes even more intensely once 80% LOAD has been hit.

Cracked floors, by default, now take 3 steps to break. If you're at 50% LOAD, they take 2 steps. 80% LOAD results in you falling instantly on the first step.

Current LOAD value is now colorized; white when below 50% LOAD, yellow when 50+%, red when 80+%.

BTC counter specifically when purchasing rings has been fixed here.

Fixed a bug where resetting your stats during allocation wouldn't actually reset END, which would allow you to infinitely level that stat as high as you wanted.


src/game.htm
```diff
@@ -5579,33 +5588,80 @@
...
-                            playSFX('floorCrackSlight');
-                            MAP[y][x] = new MapCell('crackedFloorSevere');
+                            const key = `${x},${y}`;
+                            // Determine step threshold based on carry weight
+                            const percent = getCarryWeight() / getCarryWeightLimit();
+                            let threshold = 3;
+                            if (percent >= 0.8) threshold = 1;
+                            else if (percent >= 0.5) threshold = 2;
+
+                            crackedFloorSteps[key] = (crackedFloorSteps[key] || 0) + 1;
```
packardbell95, line 5598: We could probably store these values on the tiles directly rather than in a separate object. Like maybe we could have a weight threshold, and multiples of that based on the player's load could impact that (simplified example: if the threshold for breaking is 5 units and the player steps on it with a load of 5 or more, it cracks. If the player steps on it with a load of 15 it cracks three times since 15 is 3x the load threshold)

Alternatively, we could add in a third cracked floor type too and change the tile type based on the player's weight

Currently, the minimap shows the slightly cracked floor in a thin red X while the severely cracked floor is a bold X. Maybe these should be bold regardless of the weight, or maybe we could have it respond to the player's LOAD (eg: if one step will put the player through the floor, show it as red, but maybe make it gray otherwise)

This doesn't have to change as part of this PR since this is perfectly functional now, but could be something to explore in the future

src/game.htm
```diff
@@ -9224,6 +9337,7 @@
             player.defense = baseStats.defense;
             player.strength = baseStats.strength;
             player.persuasion = baseStats.persuasion;
+            player.endurance = baseStats.endurance;
```
packardbell95, line 9340: Good catch on this

packardbell95 (approved):
Tested out the changes:
- [x] Player starts breathing once 50% of the max LOAD has been reached
- [x] Player sounds like Michael Moore at a Pizza Ranch buffet once 80% LOAD has been reached
- [x] The player can step on cracked floors...
  - [x] ...3 steps under 50% LOAD
  - [x] ...2 steps between 50% and 79% LOAD (inclusive)
  - [x] ...1 step at 80% LOAD or higher
- [x] LOAD value is now colorized
  - [x] White below 50%
  - [x] Yellow between 50% and 79%
  - [x] Red at 80%+
- [x] Buying rings reflects the player's current BTC count in the stats box
- [x] Resetting stats during allocation also includes END

I gotta say, the labored breathing sounds are hilarious. And the slow movement speed really makes it feel like the player is carrying a huge sack of garbage through the dungeon. Great work :+1:

src/game.htm
TheMilkMan6669: > Currently, the minimap shows the slightly cracked floor in a thin red X while the severely cracked floor is a bold X. Maybe these should be bold regardless of the weight, or maybe we could have it respond to the player's LOAD (eg: if one step will put the player through the floor, show it as red, but maybe make it gray otherwise)

I actually never noticed that the minimap tile changes depending on the state of the cracked floor. I'll keep that in mind

packardbell95: That might be a good indicator that they could be signaled better on the minimap, hah

Issue #77, opened 2025-06-23
Avoid Resetting the Game when Caret Browsing is Enabled


packardbell95:
# Problem
[nobo shared a recording of a runthrough he played](https://forum.agoraroad.com/index.php?profile-posts/50036/) that resulted in the game resetting on him on floor 40 when he opened a chest:

| <img src="https://github.com/user-attachments/assets/4ca26592-8f67-495e-8604-26a30894eb0f" width="300"> |
| --- |
| Upon opening a treasure chest, the game is reset |

Upon closer examination, it seems that nobo had [caret browsing enabled](https://www.microsoft.com/en-us/edge/learning-center/what-is-caret-browsing?msockid=26dec50a313e68462a51d31e306e6941&form=MA13I2) which allows a user to navigate a webpage's elements with the arrow keys on a keyboard and to click links with the Enter key. Because these keys are also used for navigating and operating menus in the game, nobo managed to unintentionally highlight and activate the RESET link which immediately obliterates the current game:

| <img src="https://github.com/user-attachments/assets/5d296f43-6381-4f86-a8eb-2dbc475ce5f6" width="300"> |
| --- |
| The moment when unintentional disaster strikes nobo |

# Solution
We should place the game options in a menu that exists within the game itself. So instead of having "Title", "Reset", and "Disable Music" on the screen as clickable links, we could have a Game Options menu that appears during gameplay if the player hits Escape. Then, in an in-game menu, they could choose an option intentionally, which would completely avoid any accidental clicks

| <img src="https://github.com/user-attachments/assets/2dccda31-dd08-421b-89d1-6d27d5f05d42" width="300"> |
| --- |
| A mockup of what a Game Settings menu might look like |

Resetting the game or returning to the title should also include at least one confirmation check to make sure that the player really does want to perform those actions. Games historically have had one or more of these checks to give the player some time to consider their actions

| <img src="https://github.com/user-attachments/assets/2e3e7cce-a925-4b4c-9b5e-83aab3d882b0" width="300"> |
| --- |
| Kirby Super Star for the SNES includes multiple prompts before allowing game data to be erased |

We could simply remove the Reset function altogether, though this may be wanted later on if we [persist gameplay in the browser's localstorage in the future](https://github.com/packardbell95/tardquest/issues/75)

PR #78, opened 2025-06-25
Move Options Into the In-Game Menu System

packardbell95:
- Removed TITLE, RESET, and DISABLE/ENABLE MUSIC links from the HTML interface
- Added a GAME SETTINGS menu that's accessible by pressing Escape
- Ending the game by either resetting or returing to the title is now blocked by confirmation menus
- Fixed a bug where persuasion attempts were not reset during mimic fights (if you fought a mimic and exhausted your persuasion attempts, then fought another mimic as the next enemy, you would start the fight with no persuasion attempts remaining)
- Added an `escapeDisabled` flag onto menus which, when enabled, will prevent a player from exiting a menu by pressing the Escape key
  - Added this flag onto the Stat Allocation menu to prevent players from abandoning their unspent points
- Added a Nobo wall that appears on Floor 40
- Code cleanup
  - Indented HTML syntax
  - Removed unused `return options;` statements

This PR closes https://github.com/packardbell95/tardquest/issues/77

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -478,6 +478,10 @@
             padding-top: 9px;
         }
 
+        .menus > p:first-child {
+            margin-top: 0;
+        }
+
         #controls {
             font-size: 14px;
             height: 130px;
@@ -509,7 +513,8 @@
             color: #f19b32;
         }
         .wall,
-        .gloryWall {
+        .gloryWall,
+        .noboWall {
             color: #e9ff00;
         }
         .floor {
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
All seems to be in working order. Love the nobo easter egg as well.

src/game.htm
packardbell95, line 1: Good catch. I didn't notice this cutoff

Issue #76, opened 2025-06-23
Persist Gameplay Setting in Localstorage


packardbell95:
Game settings should be saved in [the browser's localstorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) so that they don't have to be changed on every new playthrough. Currently this only involves disabling music, though we should provision the settings as an object to accommodate any future settings as well (eg: disabling sound effects, reducing animations, or whatever else)

This location would be distinct from gameplay data that may be introduced in https://github.com/packardbell95/tardquest/issues/75

Issue #72, opened 2025-06-21
Add fully randomized party responses


TheMilkMan6669:
Currently, when using the TALK function, a random party member will pick a random preset message to respond with in the battle log. This gets very old very fast and is a gimmick mechanic that you end up dropping immediately.

We could have it so messages are structured into "word slots". Each slot will pull from a pool of words or word groups at random and create a fully randomized message. Of course, the end result will look stupid as hell, and that's exactly what it should be. Example...

Slot 1 has these available words: "You, I, What"
Slot 2: "Are, Aren't, Can't"
Slot 3: "Gay, A total dumbass, A pot roast"
Result: "--you pull out your translator-- You are a pot roast?"

It doesn't necessarily need to be limited to 3 slots, but structurally, that might make the most sense. This would have to be played around with. Each enemy could have their own unique sets of pools to pull from as well to make their individual responses a little more unique from each other. Like the Lughead could just have a vocabulary of nonsense caveman noises or something.

To push things further, we could use something like [BetterSAM](https://imrane03.github.io/better-sam/) to vocally play these messages out. Or we could just record the words with our own voices and bitcrush the hell ouf of them, whatever ends up being preferred.

PR #80, opened 2025-06-29
New dialogue system + localstorage settings

TheMilkMan6669:
This includes the new dialogue system we discussed where when using the TALK function while you have party members, instead of pulling a random preset static message, it will instead pull random words from word/phrase pools and insert them into wordslots to create a sentence. Now you will have brilliantly nonsensical and braindead randomized dialogue such as the following:

Screenshot 2025-06-28 170659
Screenshot 2025-06-28 181621

Under the enemies object, you can easily define a specific set of word/phrase pools for specific enemies. Any that don't have their own unique set will use the default fallback set instead. I adjusted the battlelog text gradient as well (which is _not_ demonstrated in the first screenshot).

I've once again changed how the textbox is presented as well.

Screenshot 2025-06-28 183713

I've also added a toggle for level up animations via the settings menu, and implemented localstorage persistence for that and the music toggle.

Issues #76 and #72 will be left unresolved for now, because I'm sure we'll still want to add more settings to add localstorage to and we may still want to add TTS playback to the party dialogue in the future.

### DIFF
--- a/src/game.htm
+++ b/src/game.htm
@@ -449,18 +449,18 @@
         #inputBox input {
             color: white;
             font-size: 13px;
-            width: 600px;
+            width: 672px;
             height: 2px;
             z-index: 9;
             background-color: #000;
-            border: 3px dotted #A23AB4;
+            border: 2px solid #ffffff;
             padding: 10px;
             font-size: 16px;
             border-radius: 0px;
             text-align: left;
             position: absolute;
-            top: 125px;
-            left: 58px;
+            top: 117px;
+            left: 28px;
             outline: none;
         }
 
@@ -6577,6 +6577,13 @@
                 hp: 13,
                 attack: [2, 5],
                 bitcoins: 4,
+                talkSlots: [
+                    ["Sss...", "Hrk!", "Glorp", "Snail"],
+                    ["you", "I", "we", "that guy", "snail"],
+                    ["am", "is", "are", "was", "snail", "violated"],
+                    ["slimy", "slow", "fast", "hungry", "snail", "the law"],
+                    ["?", "!", "!!!", "?!", "...", ".",]
+                ]  
             },
             {
                 id: "stupidDog",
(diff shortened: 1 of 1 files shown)


src/game.htm
```diff
@@ -7257,6 +7272,23 @@
...
+            const eatRat = localStorage.getItem("eatRatAnimationEnabled");
+            if (eatRat !== null) eatRatAnimationEnabled = eatRat === "true";
+            const sfx = localStorage.getItem("sfxEnabled");
+            if (sfx !== null) sfxEnabled = sfx === "true";
+        }
+
+        function saveSettings() {
+            localStorage.setItem("musicEnabled", musicEnabled);
+            localStorage.setItem("eatRatAnimationEnabled", eatRatAnimationEnabled);
+            localStorage.setItem("sfxEnabled", sfxEnabled);
```
packardbell95, line 7287: This doesn't have to change here, but just an idea for how to handle the settings:

It might be better to save everything all at once with JSON serialization:
```suggestion
            localStorage.setItem("playerOptions", JSON.encode({
                musicEnabled, eatRatAnimationEnabled, sfxEnabled
            }));
```

To undo it would basically be something like:
```javascript
const {musicEnabled, eatRatAnimationEnabled, sfxEnabled} = {...JSON.parse(localStorage.getItem("playerSettings"))};
```

We could put all of the settings into an object too:
```javascript
const PlayerSettings = JSON.parse(localStorage.getItem("playerSettings"));
```

packardbell95 (approved):
The changes look good overall:

All of the settings are saved and loaded in the browser automatically:
- ✅ Music
- ✅ SFX
- ✅ Level Up Animation

The new dialog system is a definite upgrade as well. After dozens of interactions, some notable quotes I encountered are:
- "Glorp that guy is Judge Joody?!"
- "Halt! that guy snail fast..."
- "He can't kill the Burnout Revenge on the PS2..."
- "Listen here, you need to pick your nose!"
- "She must snack on a gay?!"

Great work all around!

PR #82, opened 2025-07-06
Animated menu + NPC portraits

TheMilkMan6669:
Whenever you enter a menu or NPC interaction, the minimap area is replaced by an animated ASCII portrait. Currently this covers the settings, menu, merchant, and gambler menus. More can be added later for the submenus.

Example video:

https://github.com/user-attachments/assets/0f65b916-cf11-4282-89c8-1654b9fa4973

I also debuffed the LOAD gain when levelling END because you really just don't get heavy often enough

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -524,6 +524,24 @@
             color: #f00;
         }
 
+        #portraitArtOverlay.portrait-art {
+            position: absolute;
+            left: 31px;
+            top: 140px;
+            width: 217px;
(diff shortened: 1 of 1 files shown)


packardbell95 (approved):
Changes here look fantastic. I absolutely love the new art. Great work!

I left some minor feedback, though it doesn't really impact anything, so I think this is fine to merge as-is

Issue #79, opened 2025-06-28
Prevent Gambler from being killable before being made visible


TheMilkMan6669:
If you place a brick of C-4 in an area where the Gambler _would_ have spawned if you had the 200+BTC prerequisite, he would be blown up anyway. This should be fixed

Issue #73, opened 2025-06-21
Update English Articles, Incorporate into Menus, Update Name Highlighting


packardbell95:
# Functional Consistency
The `getArticle()` function currently does not apply appropriately to all item types. In particular, applying it to "NOKIA MAIL" will return "a " instead of "", which is technically not grammatically correct

### ❌ Incorrect Phrasing
> You bought a NOKIA MAIL

### ✅ Correct Phrasing
> You bought NOKIA MAIL

We should check all items and equipment names to make sure that the correct article is returned, if any

# Application
Following https://github.com/packardbell95/tardquest/pull/71, we have a sale menu that does not include the article of the items/equipment being sold. We should include the article here to be consistent with purchases:
> ✅ You just bought a TORCH
> ❌ You sold TORCH for 12 BTC.

# Highlighting
Items are not highlighted consistently in various menus. These should be addressed:

| <img src="https://github.com/user-attachments/assets/7b41324a-c8da-4328-b608-afa9e86cbe16" width="300"> |
| --- |
| Names are colored when sold but not when purchased |

PR #83, opened 2025-07-06
Misc Updates

packardbell95:
This PR contains a batch of updates, including the following:
- Animation and sound updates
  - After falling into a pit, the viewport will flash red when the "splat" is heard
  - Stepping on a healing tile now flashes the screen green and plays a sound
  - Selling things now plays a kaching sound
- Player HP no longer falls below 0 HP (previously when killed by an enemy, the HUD may display a negative HP value)
- `onExplode()` is now only called for visible tiles (closes https://github.com/packardbell95/tardquest/issues/79)
- `getArticle()` has been removed and replaced with adding articles directly onto objects (closes https://github.com/packardbell95/tardquest/issues/73)
- Sales logic has been consolidated
- Persuasion updates
  - Players can now cancel persuasion/talk by hitting the escape key
  - Speech now has whitespace trimmed off of the sides of the input
  - If the player's speech is empty, random flavor text is displayed instead
- Reindented some code


packardbell95:
> There's a bug introduced with this PR that sets the default max persuasion attempt limit to 3 instead of 2. Not sure where though. Otherwise, everything else seems to be working great. If we can get that fixed we can go ahead and merge.

Is there some point of the game where you're seeing this? I can't reproduce it. The max attempt number still appears to be 2 for me

|  |
| --- |
| Exhausted persuasions after two attempts |

TheMilkMan6669:
Now I'm seeing the issue here, actually. When you actually type something, it works as intended, but if you type nothing and hit enter, it will display the random player message, but it'll treat it like the TALK function rather than PRS. I noticed this immediately when I had a party member and they responded during combat.

The second time you try to go the no typing route, it'll use up a persuasion attempt as normal. Just the first attempt is treated as a TALK function.

TheMilkMan6669:
I might have come up with a fix for the above issue, so I'll just go ahead and merge now. We should keep an eye on the persuasion attempt limit in future testing though

PR #86, opened 2025-07-08
Adding speech synthesis for party members and NPCs, misc other updates

packardbell95:
This PR changes the following:
- Adding speech synthesis for party members and NPCs
- Adding the ability to toggle speech synthesis in the Game Settings menu
- Party member names now appear `<like this>` rather than `like this:`
- Minimizing the number of places to update styles of player speech in `player.say()`
- Changing an error message from `console.log()` to `console.error()`
- Removing unneeded whitespace
- Removing unused `returnToTitleScreen` case in the Game Settings menu

### Demo
https://github.com/user-attachments/assets/a4558bbb-3ddc-4579-b341-fc8df318ab9a

### About Speech Synthesis
The speech synthesis in this update is made possible by [Imrane03/better-sam (SAM Software Automatic Mouth)](https://github.com/Imrane03/better-sam?tab=readme-ov-file#sam-software-automatic-mouth)

Some modifications to SAM have been made, including:
- Removal of all printed output
- Adding options for reverb and an `onEnded()` callback function

The game provides a wrapper for the speech synthesis via `SpeechSynthesizer`. To play any speech, you just need to supply a sentence and voice settings. For example, to say "Hello world" as the merchant, you can run:
```javascript
SpeechSynthesizer.speak("Hello world", merchant.voice);
```

Each enemy and NPC has been given a `voice` object which consists of:
- Pitch
- Speed
- Mouth
- Throat

These settings are identical to what one would use on [this Better SAM Software Automatic Mouth demo page](https://imrane03.github.io/better-sam/). It's recommended to use this page to easily test out new voice settings

The `SpeechSynthesizer` wrapper also contains an `isSpeaking` flag that is set once speech begins and is turned off as soon as playback ends (including reverb). This flag will block any other speech from playing at the same time. Simultaneous speech will still be printed in the battle log, so the player won't miss out on any information

The SAM library has been modified to include some other features as well:

#### Reverb
The reverb is a simple algorithmic implementation that adds a delay onto the speech that goes through a low pass filter. This relatively primitive implementation fits the lo-fi voices and sound effects of the game. Having these defined as options will make it easier to change the reverberated characteristics for any types of environments that we plug into the game in the future (eg: a jungle environment may have little to no reverb whatsoever)

Reverb can be set on the `options` that are passed to the `SamJs` constructor via a `reverb` object that consists of the following properties:
- `wetGain` The level of the wet signal
- `dryGain` The level of the dry signal
- `delayTimeSeconds` The duration of the delay
- `decay` The decay rate of the reverberation
- `totalDelays` How many delays are used to create the reverb effect
- `lowpassFilterCutoffHz` The frequency over which the reverb is cut off

> [!WARNING]
> It is possible to create a feedback loop of sound if the delay lasts too long or isn't cut off properly. If these are tweaked, be sure to listen for any humming or feedback noise after a set of speech has completed

#### `onEnded()` Callback Function
An `onEnded()` callback function can be passed as an option to the library that will be run once SAM has ceased playback

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@ A Javascript game by [Xx_TheMilkMan69_xX](https://forum.agoraroad.com/index.php?
 
 [endless_self](https://forum.agoraroad.com/index.php?members/endless_self.6570/)
 
+### LIBRARIES USED
+- Synthesized voices are powered by [Imrane03/better-sam (SAM Software Automatic Mouth)](https://github.com/Imrane03/better-sam)
+
 # Objective
 Fight your way to the end of the dungeon!
 
--- a/src/game.html
+++ b/src/game.html
@@ -76,9 +76,9 @@
 
         #titleScreen {
             position: fixed;
-            left: 0; 
+            left: 0;
             top: 0;
-            width: 100vw; 
+            width: 100vw;
             height: 100vh;
             background: #000;
             color: #fff;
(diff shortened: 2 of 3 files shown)


TheMilkMan6669 (approved):
Working great on my end here. The voice settings for the Snail Sentinel and Lughead are especially hilarious. I'm of the opinion that we can close #72 now that we have this implemented.

PR #88, opened 2025-07-09
Adding minimap tooltips, encapsulating portrait logic

packardbell95:
### Summary of Changes
- Encapsulating the Minimap and Portrait functionality into their own objects
- Adding tooltips to the cells on the minimap to clearly show cell details, including
  - Enlarged cell character
  - The name of the object occupying the cell
  - The cell's map coordinates
- Portraits now have colors and X/Y offsets to help center the animations
- The minimap is now hidden when a portrait is opened
- Disabling picture-in-picture browser interaction for the level up animation
- Fixing a bug where allies can join your party with 0 HP (they now always have a minimum of 1 HP)

### Minimap and Portrait Updates
This PR introduces two new objects: `Minimap` and `Portrait`. Both of these handle the drawing of each of these components on the page

#### Showing and Hiding the Minimap
```JavaScript
Minimap.show();
Minimap.hide();
```

|  |
| --- |
| A rotation of each portrait animation |

Animations are stored in the `Portrait.animations` object, keyed by the ID of the animation. The current list of animations includes:
- `chest`
- `death`
- `gambler`
- `inventory`
- `merchant`
- `settings`

For example, to show and hide the gambler animation, one would simply call:
```JavaScript
Portrait.show("gambler");
Portrait.hide();
```

Each of the `Portrait.animations` objects contains the following parameters:
- `frameDelayMs` The length of time in milliseconds that a frame will last
- `playbackMode` Either "repeat" or "once". When set to "repeat", the animation will loop forever. Otherwise, the animation will remain paused on its final frame
- `padding` An object consisting of an `x` and `y` number. These are positive integers that will add spacing to each frame to help position it within the portrait window. The `x` value is the number of space characters that are added to the left of a frame and the `y` value is the number of newlines that are inserted before the frame
- `frames` The array of frames. Each frame is a single string that is displayed in the portrait window

### Minimap Tooltips
|  |
| --- |
| Hovering over the minimap now reveals useful information |

When a cell on the minimap is hovered, it will now show the details of the cell. This includes an enlargement of the cell's character, its name, and the map coordinates where it resides

### Disabling Picture-in-Picture
Picture-in-Picture is probably the most annoying and useless modern browser feature. This PR disables it explicitly for the level up animation because apparently you can do stuff like this in Edge otherwise:
|  |
| --- |
| Literally pulling the animation out of the scene |

### 0 HP Ally Bug
|  |  |
| --- | --- |
| **Before:** Enemies with 1 HP may join with 0 HP | **After:** Enemies always join with a 1 HP minimum |

When enemies are successfully convinced to join you in your quest, they are added to the party with half of their current HP. The halving is rounded down via `floor()` which means that if an enemy has 1 HP remaining, they are technically dead when they join the team

This PR fixes the issue by ensuring that a minimum of 1 HP is set when an ally joins the team

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -545,7 +545,8 @@
         }
 
         #battleLog,
-        #minimap {
+        #minimap,
+        #portraitArtOverlay {
             background: #000;
             border: 2px solid #fff;
             padding: 10px;
@@ -561,13 +562,26 @@
             overflow-y: scroll;
         }
 
-        #minimap {
+        #minimap,
+        #portraitArtOverlay {
             height: 166px;
             width: 216px;
             padding: 2px;
             white-space: pre;
             font-size: 12px;
+        }
+
+        #minimap {
             border-top: 0;
+        }
+
+        #minimap > span {
+            cursor: help;
+        }
+
+        #minimap > span:hover {
+            color: #000;
+            background-color: #fff;
         }
 
         #minimap .crackedFloorSlight,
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Everything looks good to me. The tooltips are nicely presented and I approve of the colorized portraits. I agree we should colorize their in-game sprites as well, we can just lump that into the next update. I'll open up an issue for this one.

PR #90, opened 2025-07-11
UI Updates

packardbell95:
This PR does the following:
- Integrated gaming details into to new UI structure
- Fixed render issue after falling through a floor and dying
- Stopped NPCs from also vocalizing in addition to screaming with a sound effect when getting blown up with C4
- Integrating party member data and behavior into the `player` object
- Adding party member buttons
    - "Talk" talks to a living NPC without requiring text input
    - "Release" immediately boots an NPC out of the party
- Changed the article of the Alaska Raisins
    - "You bought/sold some Alaska Raisins..." rather than "You bought/sold Alaska Raisins..."
- Moving common code into `player.leaveCombat()`
- Added `.giveBitcoins()` and `.takeBitcoins()` helper functions to the `player` object to handle stat rerendering
- Calling `player.levelUp()` before opening the menu so that the current stats reflect the new level


TheMilkMan6669 (changes requested):
Everything seems to be in working order. I noticed however that when falling down a cracked floor, the death portrait doesn't display until _after_ the red screen effect displays (when the player hits the ground essentially). Not sure if intentional, but in my opinion the death portrait should be displaying right as the floor cracks.

Once that's taken care of, we can merge this into master and then it'll stay under the preview build on my site until we get the clickable inventory bar set up.

Issue #91, opened 2025-07-14
Add Mouse Navigation ala Ultima Underworld


packardbell95:
[Some of the UI updates](https://github.com/packardbell95/tardquest/pull/90) are introducing more mouse interactivity. However, we don't have a way to navigate the dungeon itself with the mouse. We should include clickable hotspots on the main viewport that will allow the player to navigate with a mouse

This should work similarly to Ultima Underworld where the mouse cursor turns into arrows depending on what direction the player will turn or move in:
| <img src="https://github.com/user-attachments/assets/6c61aecd-00db-42e3-975a-a1c668e61a7b" width="300"> |
| --- |
| The mouse cursor helps guide players in navigation in Ultima Underworld |

PR #93, opened 2025-07-18
Adding Mouse Navigation to the Game

packardbell95:
In an effort to integrate mouse controls further into the game, this PR introduces on-screen navigation similar to the controls in Ultima Underworld

|  |
| --- |
| Navigating the Tardspire with mouse controls 🐁 |

These changes respond to the two main game modes: navigation and combat

Navigation allows the player to mouseover and click the main viewport to move forward, backwards, turn left or right, and strafe left or right depending on whether the surrounding environment allows. Any cells directly in front of the player that have an `onTouch()` function will also change the style of the cursor, even though clicking on one of these cells will still call `move('forward')`

Combat is very simple, making the entire viewport a clickable surface. When clicking, this will result in the player attacking the enemy, basically doing the same thing as pressing the "A" button

Things like menu interactions are not covered in this PR, but will be coming in future updates along with things like more on-screen buttons to facilitate interactions

This closes https://github.com/packardbell95/tardquest/issues/91


TheMilkMan6669 (approved):
Working great on my end. The custom cursors look nice too. I'll probably see if I can come up with my own mockup or general outline for other miscellaneous clickable buttons we can use for menus and such and show it to you later.

PR #100, opened 2025-07-22
Viewport UI label + separate enemy art module

TheMilkMan6669:
Added a UI label for the viewport, as I personally think it just looks better to have one there. Also made sure the label doesn't shake alongside the viewport display when moving or turning.
Screenshot 2025-07-22 021926

I've also consolidated the enemy artwork into a separate .js module, like what was done for the portrait artwork.

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -269,6 +269,7 @@
 
         #game > .layer {
             position: absolute;
+            margin-left: 3.5px;
             top: 0;
             left: 0;
         }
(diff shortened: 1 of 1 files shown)


packardbell95 (approved):
I pushed a couple of small changes to this, the most significant of which is one that fixed mouse controls since the control handler managed to get pushed below the viewport into an unclickable area. Those aside, the changes look good :+1:

PR #101, opened 2025-07-22
Optimize Minimap Rendering Function

packardbell95:
This PR aims to provide select performance improvements for the game through restructuring the map, minimap, battle log, and menu

## Map and Minimap Updates
This PR adjusts the map's functionality to wrap it into its own `Map` class which handles a lot of placement and rendering functionality with more ease

### Updating Map Cells
Previously, map cells were tracked in the global `MAP` array. `MAP` has now been converted to an instance of the `Map` class. This means that `MAP` still holds the contents of the game's map, but now has helper methods associated with it that can be used to perform operations on the map

In particular, we previously were setting map cells by creating new `MapCell()` objects in the array's coordinates:
```javascript
MAP[y][x] = new MapCell(...);
```

Now, we can set these with `Map.setCell()`:
```javascript
MAP.setCell(x, y, type);
```

This still uses `MapCell()` behind the scenes, but performs some handling, such as map boundary checks and initialization automatically

### Rendering Optimization
Previously, on each `render()` call, the minimap would be entirely discarded and recreated. This has become notably heavy after introducing tooltips which would require a handler to be reattached to all cells after clearing

Instead of refreshing everything, the `Map` class in this PR now only updates the cells of the minimap if they had changed since the last time. When a cell is updated via `Map.setCell()`, `Map.setEntities()`, `Map.initializeMinimap()`, `Map.revealSpot()`, or `Map.reveal()`, an internal `#rerenderList` is updated with the coordinates of the changed cell. On every `render()`, `Map.refreshMinimap()` is called which will cause only the cells located in the coordinates of the `#rerenderList` to be updated rather than the entire minimap. This cuts down significantly on the number of elements that are generated and discarded, making it easier for browsers to manage the state of the minimap

### Tracking Entities on the Map
The map can also track entities apart from the map's cells via `Map.setEntities()`. This is currently only used for the player, but will be used for NPCs in the future as well. `Map.setEntities()` takes an object keyed by entity ID that contains a pair of `x` and `y` coordinates as well as cell overrides that are used to replace the display of that cell in the minimap when rendered

These objects can also have a `forceRefresh` option which, if set to `true`, will force the entity to be refreshed on every render. This is currently enabled for the player character since turning in place will not cause the cell to be updated (therefore avoiding the cell being added to the aforementioned `#rerenderList`), though the player's character is an arrow that needs to be shown pointing in the correct direction at all times. This mechanism may be replaced with something more sophisticated in the future, though given that it's only used for the player character, is a minimal concern at this point

### Map Generation Logic Organization
The `Map` class now holds the bulk of the general map generation logic. `Map.generate()` will now generate a basic map with an entry and exit point

Beyond the general map structure, the facets of the map such as placing chests and traps are handled in the game's `generateMap()` function. This is because entity placement logic is specific to TardQuest, not to a generic game

### Elimination of Random Attempts
Placing floors and cells was previously performed by selecting random cells, seeing if they are empty, and placing entities in those spots. This is incredibly inefficient since most of the cells are not going to be empty, so scanning this way was particularly excessive

Now, the `Map` class has a `.getEmptyCellCoordinates()` method. This returns a list of all of the coordinates that are empty. Combining this with a new `shuffle()` method that randomizes the order of an array now makes it much simpler to find empty spots to place entities and objects around the map

## Battle Log Changes
The battle log was creating some extra work for browsers. This is because entries to the battle log would be tracked in an array. When updated, this array would be rescanned to generate the battle log messages, styling them based on their position to give a gradient effect

This has been replaced with simple logic that just prepends new lines to the battle log element. The gradient is now styled with CSS, so elements are no longer discarded and regenerated

## Menu System Updates
The menu system has been split off into its own `Menu` class. The way that the map is rendered has been modified to be less excessive. Basic mouse support has also been added to the menus, meaning that any given menu option can be hovered and clicked

### Menu Rendering
Like the minimap, the menus were constantly redrawn on `render()` calls. This includes when a selected option within the menu has changed. Because this resulted in a lot of needless regeneration and interfered with mouse events, the menu is now rendered only once when opened. Selecting options with the arrow keys or by hovering an option with the mouse cursor will now change the selected option internally and will simply change the option's class name in HTML to `selected` which determines which option is being pointed at with an arrow ("▶")

## Cracked Floor Updates
Previously, the cracked floor would be tracked and updated within `crackedFloorSteps`

This mechanism has been simplified by introducing a third type of floor: `Cracked Floor`. This now sits in between the other two:
1. `Slightly-Cracked Floor`
2. `Cracked Floor`
3. `Severely-Cracked Floor`

The logic for how these behave is now entirely stored within each of the cells' `onEnter()` callbacks.

The `player` object also now has a `player.getWeightLevel()` function that returns `normal`, `warning`, or `danger` depending on the player's weight percentage based on their carry and endurance levels

Together, the floor now breaks based on the following logic:
- `Slightly-Cracked Floor`...
  - ...becomes `Cracked Floor` when the player's weight level is `normal`
  - ...becomes `Severely-Cracked Floor` when the player's weight level is `warning`
  - ...breaks entirely if the player's weight level is `danger`
- `Cracked Floor`...
  - ...becomes `Severely-Cracked Floor` when the player's weight level is `normal`
  - ...breaks entirely if the player's weight level is either `warning` or `danger`
- `Severely-Cracked Floor` breaks if the player's weight level is `normal`, `warning`, or `danger`

## Misc Updates
- The game can now be started by clicking on the title screen
- Party member counts are now displayed in the Party section header
- The `Tooltip` class has been split into its own file
- Warnings from Edge's developer tools, namely `meta viewport` and `-webkit-user-select` (webkit doesn't support `user-select`, strangely), have been addressed
- The player's minimap tooltip now animates the player emoticon

## Other Notes
While this PR offers some performance improvements, there are still many to make in the scene renderer itself. This is being tracked under https://github.com/packardbell95/tardquest/issues/104

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -10046,7 +10046,9 @@ A Javascript game by <a href="https://milklounge.wang/tardquest/" target="_blank
             if (awaitingPersuasionText) return;
 
             if (menu.isOpen()) {
-                menu.handleInput(key);
+                if (! animationActive) {
+                    menu.handleInput(key);
+                }
             } else if (key === 'escape' && !awaitingPersuasionText) {
                 menu.open('gameSettings');
             } else if (key === 'i') {
--- a/src/game.html
+++ b/src/game.html
@@ -606,6 +606,7 @@
         }
 
         #minimap .crackedFloorSlight,
+        #minimap .crackedFloor,
         #minimap .crackedFloorSevere {
             color: #f00;
         }
@@ -945,6 +946,7 @@
         }
 
         .scene_character_crackedFloorSlight,
+        .scene_character_crackedFloor,
         .scene_character_crackedFloorSevere {
             color: #000;
         }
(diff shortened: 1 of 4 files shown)


TheMilkMan6669 (approved):
Changes all seem good. The additional mouse input features are very nice QOL changes, and having a third cracked floor sprite is nice and was something that was definitely needed. 

The only small issues I came up with are there's currently no way (from what I could tell) to use the mouse to scroll menus that have multiple pages, and clicking on one of the dev name links results in the game starting when the player might possibly not want it to yet. We could remedy this by putting a play button somewhere on the title screen instead of having the entire screen be a clickable "surface", but we could handle this in a future update. It's really no big deal.

packardbell95:
> The only small issues I came up with are there's currently no way (from what I could tell) to use the mouse to scroll menus that have multiple pages, and clicking on one of the dev name links results in the game starting when the player might possibly not want it to yet. We could remedy this by putting a play button somewhere on the title screen instead of having the entire screen be a clickable "surface", but we could handle this in a future update. It's really no big deal.

Good call-outs on these. I think they're worth implementing. These are now tracking under https://github.com/packardbell95/tardquest/issues/105

PR #106, opened 2025-08-03
Ring of Sightliness Changes + Equipment menu combat limit

### LINKED ISSUE #65: Stop Ring of Sightliness from Seeing Through Walls
When the player has the Ring of Sightliness equipped, they can see through walls on the minimap. This is because the ring uncovers a radius around the player with `revealMapSpot()` which will uncover the map in a circle, regardless of what's beneath

We should adjust the function to limit the visibility to stop at walls. This should be an optional parameter used for player movement since this function is also used for the Dowsing Rod to reveal the exit and its surrounding map cells. Wall obstruction doesn't need to be accommodated in this case

TheMilkMan6669:
A fairly simple set of small changes. The Ring of Sightliness will no longer allow you to see directly through walls, and basically just increases the player's FOV cone. Below is an annotated (maybe somewhat kinda inaccurate) screenshot of what the player's FOV cone looks like after equipping the ring, and what tiles are being revealed.
fov

The equipment menu is also now inaccessible during combat, which was a decision made to reflect the upcoming clickable UI we have planned, complete with a bit of flavor text in the battle log.
battle-restriction

This PR will close #65

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -8743,9 +8743,32 @@ A Javascript game by <a href="https://milklounge.wang/tardquest/" target="_blank
         }
 
         function updateSeenTiles() {
-             // Default reveal radius 1, 2 if Ring of Sightliness equipped
+            // Default reveal radius 1, 2 if Ring of Sightliness equipped
             const radius = player.hasEquippedRing('ringOfSightliness') ? 2 : 1;
-            MAP.revealSpot(player.x, player.y, radius);
+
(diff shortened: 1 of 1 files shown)


packardbell95 (approved):
This is an awesome update. The new FOV check works really well and gives a more realistic feeling to exploration

|  |
| --- |
| Using an exaggerated radius shows the realism of the sight cone |

The new vision cone makes me want to go play some [C-Dogs](https://dosgames.com/game/c-dogs/)

Issue #95, opened 2025-07-20
Party Member Interactions Sometimes Stay Disabled


packardbell95:
Sometimes, the party member interactions will remain disabled and can't be clicked. This was observed after a team member was eviscerated after a battle where persuasion was used, which may have interfered with the buttons

| <img width="300" src="https://github.com/user-attachments/assets/957b6acd-f684-4ff9-baff-3bbe00202dc9" /> |
| --- |
| Disabled ally buttons post-fight |

Issue #105, opened 2025-07-30
Adjust Mouse Interactions


packardbell95:
From MilkMan's PR feedback:

> The only small issues I came up with are there's currently no way (from what I could tell) to use the mouse to scroll menus that have multiple pages, and clicking on one of the dev name links results in the game starting when the player might possibly not want it to yet. We could remedy this by putting a play button somewhere on the title screen instead of having the entire screen be a clickable "surface", but we could handle this in a future update. It's really no big deal.

_Originally posted by @TheMilkMan6669 in https://github.com/packardbell95/tardquest/pull/101#pullrequestreview-3069941816_

PR #107, opened 2025-08-05
Adding button inputs on UI, other changes

packardbell95:
- Adding MilkMan's UI art assets (note that inventory assets are not yet in use within this PR, but buttons are)
- Adding input section for clickable controls
- Reducing clickable area on title screen to the 'Press Enter' box
- Adding description for menu interactions in the keyboard controls section
- Entering empty persuasion text now uses an empty message so the player no longer wastes a turn in combat
- Changing MouseControl to GameControl since it now handles more on-screen controls
- Setting and fixing onOpen() and onPageChange() functions to the Menu class
- Adding getPagination(), getTotalPages(), and pagination helpers to the Menu class
- Moving death check to player-adjusting functions
- Adding numberBetween() and randomEntry() helper functions
- Adding/fixing colors for select battle log messages
- Fixing screen shake interger calculation
- Adding Confirm and Close buttons to the persuasion text input
- Adding more flavor text
- Simplifying logic

Closes https://github.com/packardbell95/tardquest/issues/95 and https://github.com/packardbell95/tardquest/issues/105


TheMilkMan6669 (changes requested):
Very nice set of updates, though I'd suggest we wait until we get the inventory icons in there before we merge. The PRS icon currently has a green highlight, I would suggest making this purple instead to match with the actual color of the PRS stat. Run could probably be blue or something.

Edit: I've also found that if you select the "back" option using the confirm icon, the up/down icons become greyed out, but not if you hit escape to go back a menu.

packardbell95:
@TheMilkMan6669 Thanks for the quick and thorough feedback

> The PRS icon currently has a green highlight, I would suggest making this purple instead to match with the actual color of the PRS stat. Run could probably be blue or something.

Great call. I've updated the colors so that attack is red, persuasion is the same purple as the persuasion stat color on the left, and run is the default blue:
|  |
| --- |
| I'm about to deal this stupid dog a rainbow of pain |

> Edit: I've also found that if you select the "back" option using the confirm icon, the up/down icons become greyed out, but not if you hit escape to go back a menu.
> 
> Edit 2: Sometimes this happens when hitting escape too actually, I can't exactly find a consistent way to replicate the issue. But it does happen lol

Good catch on this. I posted a fix

Some details about why this happens:
This PR has [an object called `GameControl`](https://github.com/packardbell95/tardquest/blob/94f4b25ab04b6a1d13115647f386e096b5eadf3a/src/game.html#L7092) which handles most of the clickable on-screen elements. It works in different modes that we shouldn't have to really think about. So long as we call [`GameControl.update()`](https://github.com/packardbell95/tardquest/blob/94f4b25ab04b6a1d13115647f386e096b5eadf3a/src/game.html#L7178), it will figure out if the mode has changed, display the correct buttons if the mode has switched, and then enable/disable the appropriate buttons

So for Navigation mode, directions that the player can't move in due to walls will disable the corresponding buttons. In battle mode, it will disable Persuasion if the player has run out of attempts (though in its current state, this can still be activated with the keyboard which might be something for us to change). And in menu mode, it will disable up and down if there aren't multiple options, and will disable pagination if there are no pages to turn to

To get this to work with the menu system, the menu has a series of callbacks that fire under different events, such as when the menu is opened, when a page is turned, when a selection changes, and so on. [We set these right after the menu has been initialized.](https://github.com/packardbell95/tardquest/blob/94f4b25ab04b6a1d13115647f386e096b5eadf3a/src/game.html#L10651-L10659) This has made it possible for us to do things like play selection sounds as the player navigates through options and pages. But now there are a bunch of callbacks to fire `GameControl.update()` in different situations

The problem was that `GameControl.update()` was not being fired on select or on cancel

I was able to replicate the issue by trying to sell items to the merchant while having enough items in my inventory to push the "[Back]" option onto the second page by itself:
|  |
| --- |
| The back button, all alone, disabling the buttons in the bottom right corner of the UI |

Selecting the option [fires the menu's `onCancel()` callback](https://github.com/packardbell95/tardquest/blob/94f4b25ab04b6a1d13115647f386e096b5eadf3a/src/scripts/Menu.js#L196) if one has been set (which fires when a menu is backed out of. [`onClose()` is fired instead](https://github.com/packardbell95/tardquest/blob/94f4b25ab04b6a1d13115647f386e096b5eadf3a/src/scripts/Menu.js#L194) if the menu system is closed completely)

Basically, I had no `GameControl.update()` call in the `onCancel()` event, but [I plugged that in](https://github.com/packardbell95/tardquest/pull/107/commits/94f4b25ab04b6a1d13115647f386e096b5eadf3a#diff-add1bebb6aadf116dc859e7ae4248fff45ff56f4532408bf8f13c282acdda6d9R10658) and have not been able to reproduce the issue since

PR #113, opened 2025-08-11
Enhanced speech synthesis with pronunciation overrides and number handling

### LINKED ISSUE #110: Support Number Pronunciation for Multiple Digits in Speech Synthesizer
[SamJs](https://github.com/packardbell95/tardquest/blob/32c2e7cc5fa2bcb0a61dd302d5e9b690ef4a54e2/src/scripts/sam.js) does not pronounce entire numbers

For example, when trying to sell an equipped weapon to the merchant, the merchant says "You sure you wanna point that thing at me? I've got a 12 gauge hidden under my sleeve, you know..."

When spoken with the speech synthesizer, this pronounces "12" as "one-two" instead of "twelve"

### Reproduction Steps
An easy way to test this is to open the browser's Developer Tools, then paste this into the console:
```javascript
merchant.say("12");
```
This will synthesize the speech "one two" instead of "twelve"

### Possible Implementation
The fix could take place either in [the SpeechSynthesizer object](https://github.com/packardbell95/tardquest/blob/32c2e7cc5fa2bcb0a61dd302d5e9b690ef4a54e2/src/game.html#L6450) or in [the speech library](https://github.com/packardbell95/tardquest/blob/32c2e7cc5fa2bcb0a61dd302d5e9b690ef4a54e2/src/scripts/sam.js) directly

There could be a preprocessor in either of these that extracts groups of numbers. both unformatted (eg: "1234567") or formatted (eg: "1,234,567" or "1.234.567") and translates them into a full, pronounceable text ("One million, two hundred thirty four thousand, five hundred and sixty seven" or a phonetic equivalent that works with the library)

### No-Gos
The speech synthesizer mispronounces a lot of other words as well. Using the same example from above, "gauge" is pronounced as "gog". But this goes beyond number grouping, so this could be saved for another issue if it starts to become distracting

VocaPepper:
### Changes

1. **Pronunciation Overrides**  
   - Added `pronunciationOverrides` dictionary for irregular words (e.g., "gauge" → "gayge")
   - Implemented case-preserving substitution in `applyPronunciationOverrides()`
   - Handles uppercase acronyms and title case words appropriately

2. **Acronym Expansion**  
   - `expandAcronyms()` converts letter groups to spaced letters ("EXP" → "E X P"). **_This may require further testing._**
   - Skips overridden terms and alphanumeric mixes
   - Preserves apostrophes in possessives

3. **Numeric Processing**  
   Three-stage normalization in `normalizeNumbers()`:
   - **Fractions**: `(a/b)` → verbalized ("3/4" → "three fourths")
     - Handles signs, reduces fractions, and supports large denominators
   - **Decimals**: `x.y` → verbalized ("12.34" → "twelve point three four")
     - Processes comma-separated integer parts
   - **Integers**: Full verbalization with sign support ("-1,024" → "negative one thousand twenty-four")

The changes maintain backward compatibility while significantly improving speech output quality.  

Fixes #110

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -6449,17 +6449,40 @@ A Javascript game by <a href="https://milklounge.wang/tardquest/" target="_blank
 
         const SpeechSynthesizer = {
             isSpeaking: false,
+            pronunciationOverrides: {
+                "gauge": "gayge",
+                "colonel": "kernel",
+                "queue": "cue",
+                "yacht": "yot",
+                "epitome": "eh pit oh mee",
+                "faq": "eff ay cue",
+                "jpeg": "jay peg",
+                "gif": "gif", // pronunciation debate
+                "regex": "rej ex",
+                "cache": "cash",
+                "gaol": "jail",
+                "ache": "ake",
+                "subtle": "suttle",
+                "victuals": "vittles",
+                "often": "offen",
+                "solder": "sodder",
+                "suite": "sweet",
+                "gouge": "gowj",
+            },
+
             speak: (sentence, voiceOptions, onEnded) => {
-                // Don't speak a new line if one is currently playing
                 if (!speechEnabled || SpeechSynthesizer.isSpeaking) {
                     onEnded?.();
                     return;
                 }
-
                 if (!voiceOptions) {
                     console.error("No voice options provided", { sentence });
                     return;
                 }
+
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Seems to be working as intended. I went ahead and added a few more word overrides.

PR #120, opened 2025-08-17
Current music + Gamepad connection status bar

TheMilkMan6669:
There are now two thin boxes at the bottom of the interface displaying the current song being played, as well as a status indicator for whether or not a gamepad is connected.

Currently I am having issues actually aligning these boxes with the rest of the interface, help will be needed here, as I am seemingly lacking understanding of this whole interface grid system that's been semi-recently devised. 
(fixed by @packardbell95)

You can also now use the diceroll function during level ups, and there is even an option in the settings menu that allows you to auto diceroll and skip the level up/allocation screens entirely if you really hate yourself.

I've also prefixed the rest of the instances of BTC with ₿ while I was at it.


VocaPepper:
> That said, I have one question about the gamepad buttons: are the Talk and Inventory buttons swapped in the UI? The Navigation Controls section lists the X button as Talk and the Y button as the Inventory:
> 
> 
> Talk is labeled as the X button, Inventory is listed as the Y button
> Yet my two X-Box-related controllers have the buttons swapped:
> 
> 


I retested between my Dualsense 5 and Xbox One S controller and the mapping is correct (Triangle/Y = Inventory and Square/X = Talk). Maybe an issue with button mapping on older controllers?

PR #128, opened 2025-08-20
When an enemy is defeated they will now turn into a pile of pixels

VocaPepper:
### Adds a new animation when an enemy is defeated:

anim-demo
c4-demo

### Other Notes
- Fixed player being able to attack during PlayerDamaged animation
- `endOfPlayerTurn` was modified for this but appears to be functioning as intended.
- Tested on LibreWolf 135.0.1-1 and Microsoft Edge 139.0.3405.102

> I really, really need to see the enemy sprite die or turn into a pile of pixels or something before the curtain drop. - @packardbell95


packardbell95 (approved):
Needless to say, this is awesome

The only thing I noticed is that the collapse animation seems a touch slow. [I opened a follow-up PR to speed it up,](https://github.com/packardbell95/tardquest/pull/129) so please give that a look and let me know if it feels more satisfying that way

PR #131, opened 2025-08-24
Passive and Rolling Level Ups + More Cheat Commands

VocaPepper:
### Main Changes:
This update allows for scenarios where the player could level up more than once, such as a boss fight.

- `refreshStats()` will check if player has enough EXP to level up, and if the stat allocation menu is already open before triggering another allocation.
- Prevented the rat animation from playing if a menu is open (e.g., during rolling-level up).
- Added additional cheat commands.

### New Cheats:
- `/giveexp <amount>` Adds EXP to player.
- `/k` Kill yourself.

2025-08-2320-16-35-ezgif com-video-to-gif-converter

### TODO:
- [ ] Update VocaGuard to support rolling level up scenarios and explicitly detect when cheat commands are used. <sup>*This update can still go live as this is currently not an issue.</sup>

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -9099,6 +9099,33 @@ A Javascript game by <a href="https://milklounge.wang/tardquest/" target="_blank
                 $luck.innerText = stats.luck;
 
                 player.party.refresh();
+
+                // Level up if we have enough experience outside of combat
+                if (
+                    !player.inCombat &&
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Looks and functions good to me, I will go ahead and merge.

PR #133, opened 2025-08-25
Dog NPC

TheMilkMan6669:
A dog named Erok will now spawn on every floor (unless killed, but you wouldn't do that, would you?). Interacting will open a menu where you can pet the dog. He will run away when you leave. That is all.

https://github.com/user-attachments/assets/aea25138-f56d-4a23-8d18-373e14d7c5a4

This whole NPC is an inside joke with another one of my friends. Deal with it.


TheMilkMan6669:
Ignore the fact that I forgot to fix the "now playing" music bar for this specific interaction, I will fix this later

VocaPepper (approved):
Everything looks good with the fixes applied.

Issue #35, opened 2025-05-07
Add reaper type enemy


TheMilkMan6669:
Similar to the one found in Persona 3-5. If the player spends too much time on one floor (naturally, by taking to many steps forward/backward), a The Grim Reapist spawns. It should have a fixed set of ridiculously high stats, it should not scale based on the floor it spawns on. Only a high level lategame player can even attempt to fight them. Extremely high risk, but extremely high reward with how much EXP and BTC they drop when defeated. It should be visible on the minimap, and it will move a space on the map whenever the player moves.
This will prevent the player from spending TOO much time grinding on one floor, while also giving a very nice reward if the player does manage to survive long enough to stand a chance at fighting one.
If we wanna make them EXTRA scary, we could have it so they are able to destroy walls to beeline your general direction. This would make C-4 an excellent escape tool, thus giving that item even more utility.

packardbell95:
This is an awesome idea

What happens if the Reaper is defeated? Maybe it comes back more aggressively each time until the player leaves? Stay dead for the remainder of the floor?

TheMilkMan6669:
In my opinion, just keep him dead until you exit the floor. Additionally,  if you successfully run away, he'll stay stuck in place for a couple steps before he starts chasing again

TheMilkMan6669:
> In my opinion, just keep him dead until you exit the floor. Additionally,  if you successfully run away while fighting, he'll stay stuck in place for a couple steps before he starts chasing again

Due to an inside joke between a relative and I, we are now turning the Grim Reapist into the Gay Cocksucking Vampire, and when he is encountered,  you will be met wity a battle log message stating that he has come to suck you down.

Issue #125, opened 2025-08-19
Audio Cleanup: Convert WAV to a Lossy Format, Remove Unused Tracks


packardbell95:
We currently have a number of sound effects and even musical tracks that are stored as lossless WAV files. Converting to MP3s would drastically reduce the filesizes, meaning that players could get started playing the game sooner, or could reduce network times in case a piece of audio is loaded later on

For instance, the Title track `title.wav` is 2.4 MB and only 13 seconds long. Meanwhile, the Gambler's Theme is 2.1 MB and 1:31 long, which is 7 times longer in audio length and still 0.3 MB smaller than the theme

We also have unused tracks such as `explore.wav` which has since been converted into `explore1.mp3`. Since `explore.wav` is no longer used, it can be removed from the codebase. We should look for any other instances of unused audio and remove them as well

### A Note About Looped Audio
When a piece of music is converted to MP3, it should be tested in-game to ensure that it loops properly. Some platforms have issues with looping audio in HTML5 and may leave audible gaps between the ends of the track. This tends to happen with uncommon encoding methods, such as anything lower than 128 kbps or anything other than 44.1/48kHz sample rates

packardbell95:
@VocaPepper That's a good point. [OGG is widely supported across pretty much all systems and browsers.](https://caniuse.com/ogg-vorbis) I've changed the title of this issue to not be format-specific so we can compare encoding methods

We may need to recover any original WAV files or encodings of files that have already made it to MP3 if we want to make the formats uniform (re-encoding MP3s as OGG directly probably wouldn't be great for the audio quality). Though file format uniformity isn't as much of a goal as simply getting away from using heavy formats like WAV

packardbell95:
@VocaPepper I see [on the `preloading-concept` branch](https://github.com/packardbell95/tardquest/compare/master...preloading-concept#diff-b53f110b72a0c8d1b0c33b42478de6b3fc120e2bd2d7bde7cf2b4102f973eb2f) that some of the converted files are increasing the file sizes, such as `ash.mp3` -> `ash.ogg` increasing by 2.2 KB, or even `footsteps.wav` -> `footsteps.ogg` increasing by nearly 1 KB (which is a bit unexpected coming from WAV, though the sound is quite short)

Have you experimented with comparable quality or bitrate settings when re-encoding these by chance?

PR #136, opened 2025-08-26
Adding the Vampire

packardbell95:
This PR introduces the vampire. If the player spends too long on a floor, the vampire will spawn and chase down the player. If the player manages to defeat him, he will reset for a few rounds, then repeat the chase. With every defeat, the vampire will respawn faster, chase down the player faster, and will be harder to defeat. The only way to avoid the vampire is to exit the floor

The number of steps the player needs to take on a floor to make the vampire spawn are twice the number of empty (`! isSolid`) cells that are on the map

This closes https://github.com/packardbell95/tardquest/issues/35, but with no high reward. This is because the vampire can't truly be killed. It is a force avoid, not one to overcome

### Spawn The Vampire
If you want to spawn the vampire, just run this in the console, then take a step:
```javascript
vampire.playerStepsUntilActive = 1;
```

All vampire graphics and the vampire battle theme are by @TheMilkMan6669

This PR also introduces a simple test runner which currently adds some test coverage for the Map class. Tests are run just by opening `src/tests/TardTest.html`
> [!NOTE]
> Everything in `src/tests/` is meant only for dev testing. This folder should not be included when deploying to production

### Known Issues
We can probably tweak these later:
- Entering a battle or opening a menu when encountering the vampire at the same time can cause the events to conflict
- Some of the new sounds (specifically the bat wing flap sounds when the vampire is moving) aren't great
- The vampire doesn't respond to being detonated with C4


packardbell95:
> nice, OGG Vorbis. I ended up switching back to mp3 because I was having issues with short sound effects becoming larger in file size. I'll have to adjust my draft PR that referenced #125
> 
> edit: another thing I'm bringing in is metadata on the tracks Screenshot 2025-08-26 001706

The metadata is a good idea. Are you just including artist name and track title for now?

packardbell95:
[The latest commit](https://github.com/packardbell95/tardquest/pull/136/commits/156cb789d395257b8ae9c21be8092888013bb755) includes several changes:
- Rebalanced vampire stats
- Fixed a bug where enemy attack stats were not being used
- Fixed a discrepancy with load limit between when the breathing starts and when the load bar starts flashing
- Killing Erok now carries a luck penalty
- The gambler has an alternative greeting when you are wearing his ring
- Fixed a bug where the Erok menu could not be closed with the Escape key
- After stepping on a tile that opens a menu on the same step that spawns the vampire and closing the menu, the focal point now correctly targets the vampire's location instead of pointing to the top left corner of the screen

One of the most notable of these is the attack bug. All enemies have their own ranges of damage contained in each enemy's `attack` array, but this was being ignored. Instead, all enemies have been using the same range of 1 to 5 points. Since this bug has been around for well over a month, I've adjusted all of the existing enemy stats to have this same range since the existing ranges make the battles much harder

PR #137, opened 2025-08-27
Content centering

TheMilkMan6669:
Included here is an index.html that will display the game in its full glory centered in the middle of the window. It will also dynamically zoom the content and fill the window's height. This will especially look nice if I/we ever decided to add artwork borders here. I've also removed all padding from the area surrounding the content in game.html as well.

https://github.com/user-attachments/assets/270b4457-d12c-4020-a1e8-14a46ab6833f


packardbell95 (changes requested):
The rescaling is pretty smooth. Though I did spot one issue: when the window is horizontally narrowed, the right side of the UI starts getting cut off:

|  |
| --- |
| Horizontal shrinking resulting in UI cropping |

TheMilkMan6669:
Working perfectly on my end here, I think we can safely merge now.

Issue #123, opened 2025-08-19
Torch Glow Overlaps Viewport Header Text


packardbell95:
When the torch is active, the screen glow overlaps the "Viewport" UI header

| <img src="https://github.com/user-attachments/assets/0d96274d-9dbb-4cab-8e9e-4dd92ab05740" width="300"> | <img src="https://github.com/user-attachments/assets/52e65389-ccac-497b-b999-a05ace853cbe" width="300"> |
| --- | --- |
| The torch active in the UI | Closeup of the Viewport title text where the overlapping is visible |

### Steps to Reproduce
1. Use a torch
2. Look at the Viewport title text
   - **Expected** The Viewport title text does not glow
   - **Actual** The Viewport title text is glowing

PR #141, opened 2025-09-05
Preventing Viewport Overflow

packardbell95:
When holding a torch, the glow encroaches on the Viewport header. This PR stops that from happening

Closes https://github.com/packardbell95/tardquest/issues/123

|  |  |
| --- | --- |
| **Before:** Torch glow encroaching on the Viewport header | **After:** The glow is contained |


TheMilkMan6669 (approved):
Working well on my end, nothing else I can really add here lol

Issue #97, opened 2025-07-20
Call Out Party Slots


packardbell95:
We don't have anywhere where we call out the number of party slots. Players are expected to just know that if the Party section has three allies that they've reached the maximum before they try persuading again

Some ideas to communicate this better include putting the number of slots in the Party header (eg: "2/3" when two slots are occupied) or having empty visual placeholders that get filled in with names and stats when a party member joins:
| <img width="300" src="https://github.com/user-attachments/assets/610ba720-74bf-4b58-a608-946826c7eb2c" /> |
| --- |
| Listing the number of party members in the header and including placeholder slots could better communicate the party limit to the player |

TheMilkMan6669:
This would be a good idea, it'd certainly look nicer than the text we have there right now currently.

PR #143, opened 2025-09-06
Updating the Party section

packardbell95:
Rather than have placeholder text in the Party section at the bottom of the UI, we should have placeholder slots to more obviously communicate to the player that they can have up to 3 members in their party

|  |
| --- |
| Placeholder slots in the Party section |

Closes https://github.com/packardbell95/tardquest/issues/97


TheMilkMan6669 (approved):
Works well on my end and the section looks much nicer, good work

PR #144, opened 2025-09-06
Pigeon NPC

TheMilkMan6669:
This PR will add NPC functionality to the pigeon framework once ready (note: this branch is based off the [pigeon-framework](https://github.com/packardbell95/tardquest/tree/pigeon-framework) branch).


packardbell95 (changes requested):
There's a lot of good progress here. The pigeon's portrait is excellent too

There are a few issues I've noticed that should be addressed. I'll tag important things with "⚠️". Other feedback can be saved for later:

### ⚠️ Destructive Merge Conflict Resolutions
`master` was recently merged into this branch which caused some conflicts. The resolution resulted in changes being deleted such as [the recent party section updates](https://github.com/packardbell95/tardquest/pull/143), and perhaps others. Those should be restored

If either of you need help resolving a conflict in the future, hit me up before committing. It's generally a lot easier to resolve a conflict and preserve both changes than it is to commit to deleting changes and then have to re-add them later. Plus it preserves the origins of the code accurately through proper merge resolution

### Pigeon Mechanism Feedback
These are some things with the pigeon mechanism specifically:

#### ⚠️ Pigeon Persistence after Receiving a Message
Instead of delivering one message per floor, the pigeon will follow the player and keep delivering new messages

#### ⚠️ Persuasion Bug
The pigeon interaction hijacks the persuasion input text box which doesn't get reset after interacting with the pigeon. So when trying to persuade an enemy during battle, the game thinks you're trying to send pigeon messages

|  |
| --- |
| Using Persuade during battle opens the pigeon message input instead |

#### Server Message Truncation
The server seems to cap the input message length at 420 characters (dude weed lmao). The client input box should also cap the length at the same length to avoid any unintended truncation

#### Too Easy to Trigger HTTP 429: "Too Many Requests" During Purchase
If you try to buy too many pigeons at once (and why wouldn't you? They're free), it's really easy to trigger an HTTP 429: Too Many Requests response on the server. One way to solve this would be to limit the transactions to one per floor. Another would be to have the pigeon just roam on its own, and if it's encountered, maybe it could deliver a message, so the free transaction wouldn't even come into play:

|  |
| --- |
| "Too many requests" response from the server when trying to buy a lot of pigeons at once |

#### Publishing Server Code
It would be cool to have the server code so that this could be tested and interacted with during development as well. Right now, it's a black box and we don't have a way to test the server itself, so changes around pigeon interactions will be flaky unless they're implemented only by @VocaPepper

VocaPepper:
> It would be cool to have the server code so that this could be tested and interacted with during development as well. Right now, it's a black box and we don't have a way to test the server itself, so changes around pigeon interactions will be flaky unless they're implemented only by @VocaPepper

This is something that’s a work in progress. At this moment, the server code cannot be released until I strip some stuff down and provide instructions on setting it up in a dev environment. I won’t really have steps for deploying to prod since I doubt many people run Flask apps on IIS (it’s also a pain in the ass). I’m going to be revisiting TardBoard as well since currently the Pigeon system relies on the VocaGuard session provided by it, which is not ideal if TardBoard fails to load.

> Too Easy to Trigger HTTP 429: "Too Many Requests" During Purchase

Free pricing is for development, will be set between 5-10 I believe. Server hard caps how many you can buy per session, which can be adjusted super easy using the Flask limiter.
Right now, it compares local count and server count in case a player attempts duping and blocks the usage. 

> The pigeon interaction hijacks the persuasion input text box which doesn't get reset after interacting with the pigeon

Should be a simple fix. Didn’t even cross my mind to test this lol

> Destructive Merge Conflict Resolutions

I did not get notified of any merge conflicts when I pulled from master, so this is unexpected. Usually VS Code will flag conflicts if they exist before merging.

TheMilkMan6669:
> I did not get notified of any merge conflicts when I pulled from master, so this is unexpected. Usually VS Code will flag conflicts if they exist before merging.
Edit: I actually don’t know if this was me or not

That was most likely my fault, my apolocheese

VocaPepper:
Purchase/usage limits have been relaxed for testing (10 -> 69 <sup>nice</sup>)

Issue #142, opened 2025-09-05
Fix inventory tooltips persisting on screen


TheMilkMan6669:
Sometimes, the tooltip for items on the inventory sidebar will persist on-screen forever. This most notably happens when you run out of a specific item, like when you've drank all your lean or used up all your torches.

PR #145, opened 2025-09-06
Deleting tooltips when the sidebar button exhausts an item

packardbell95:
When using items in the sidebar, an item's button in the sidebar will be removed when the player no longer has any more. However, any tooltip reference that still exists will not be freed, so the tooltip on the page will persist since nothing exists to dismiss it

This PR fixes the problem by deleting the tooltip reference if the last of an item was used

|  |
| --- |
| Tooltips now being cleared after using up items |

Closes https://github.com/packardbell95/tardquest/issues/142


TheMilkMan6669 (approved):
Thanks for handling this one, will merge now.

Issue #146, opened 2025-09-07
Fix Carrier Pigeon utterly demolishing everything in its path


TheMilkMan6669:
Currently, the Carrier Pigeon will essentially delete any other entities (merchants, gamblers, etc) in it's path as he makes his way to the player to deliver his message. He must be stopped.

PR #147, opened 2025-09-08
Moving the carrier pigeon to an entity instead of a grid cell

packardbell95:
### Overview
The current pigeon moves by replacing map cells. This causes any cells that the pigeon flies over to be cleared since each is replaced by a floor cell

This PR fixes the issue by adding the pigeon as one of the `GameMap`'s `entities` instead of having it manage its own tile space on the map directly. Changes here also remove references to the notion that the pigeon is a unique entity since there are many of them

Closes https://github.com/packardbell95/tardquest/issues/146

### Screen Capture
|  |
| --- |
| The pigeon moving through the map without overwriting cells, except for when it dies |

### Testing
I tested this by doing the following:
1. Run `localStorage.setItem("pigeonPendingMessage", "ayy lmao")`
2. Reload the page
3. Start the game
4. The pigeon is not visible
5. Light a torch to see the map
6. Observe the pigeon on the minimap
7. Make steps to move the pigeon
8. Observe the pigeon moving on the minimap, not overwriting anything that it passes over
9. Interact with the pigeon
10. Observe that the message can be read
11. Repeat steps 1 through 8
12. Blow up the pigeon
13. Observe that the pigeon leaves a splattered corpse behind (unless it's over a non-floor, like a treasure chest)

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -6514,10 +6514,10 @@ A Javascript game by <a href="https://milklounge.wang/tardquest/" target="_blank
                         );
 
                         // NPC check
-                        // Only the vampire is here for now until we get the
-                        // other NPCs organized together
                         if (vampire.isAt(tx, ty)) {
                             gridCell.push({ type: "vampire" });
+                        } else if (pigeon.isAt(tx, ty)) {
+                            gridCell.push({ type: "pigeon" });
                         }
 
                         row.push(gridCell);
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
This is working very well on my end, thanks for pacifying the evil scourge that was our former Carrier Pigeon. See you in hell, foul beast.

VocaPepper:
> This might be something for @VocaPepper later, but it would be cool to delete the active message from `localstorage` if the pigeon is killed on a floor, similar to having read the message. Otherwise, the next pigeon will deliver the same as that first message, as if it hadn't been destroyed

In theory we can just clear it the same way the `displayDeliveredMessage()` function does in `pigeon.js`, but I'll test this later.

PR #153, opened 2025-09-12
Better Client API Script Documentation

VocaPepper:
This PR simply updates documentation for `tardboard.js` and `pigeon.js` and removes a redundant VocaGuard update call.


src/scripts/pigeon.js
```diff
@@ -1,26 +1,69 @@
...
+ * This module implements a messaging system that allows players to send and receive
+ * messages using virtual carrier pigeons. Messages are sent to a remote API and
+ * delivered asynchronously. The system includes:
+ * 
+ * - Message composition and sending
+ * - Automatic polling for incoming messages
+ * - LocalStorage persistence for pending messages
+ * - Session-based authentication via VocaGuard
+ * 
+ * @author VocaPepper
```
packardbell95, line 13: `@author` tags might be redundant since the authorship is already within the commit data

VocaPepper: Fair point!

PR #156, opened 2025-09-13
Settings button repositioning + Merchant spawn changes

TheMilkMan6669:
The positioning of the clickable settings button has been moved, and the TardPad indicator has also seen some changes. Clicking the TardPad icon will bring up a modal telling you your gamepad connection status, as well as the device name. It also makes it a little more clear to the player that you need to press a button first to activate the controller in-game.

> 

> 

> 

I did plan on using colorized gamepad icons depending on whether a gamepad is connected or not, like how we had the red/green text previously, but I wasn't able to get something working for that. Maybe @VocaPepper might be able to look into this. I kept the icons included under `/assets/interface`

And also, there is now a limit in place for how many floors you can go without a Merchant spawning, as it was very common to go a very long time without one ever spawning. This is currently configured to have him be force spawned after 3 floors of his absence, meaning if you clear 3 floors without a Merchant spawning, he will spawn on the 4th floor.

## TO-DO before setting this PR as ready for review:
- Dynamically change the colorized TardPad icons depending on connection status


packardbell95:
This is looking good so far. A couple of initial thoughts:

### Colorized Icon Ideas

> I did plan on using colorized gamepad icons depending on whether a gamepad is connected or not, like how we had the red/green text previously, but I wasn't able to get something working for that

One way to do this would be to add a `<div>` that's the same size as the icon. Give it a `position: absolute` and place it over the icon, then set its `mix-blend-mode` to `darken`

I tried this by giving `#controllerStatusButtonContainer` a `position: relative` rule, then creating a `<div>` as a sibling to `#controllerStatusButton` with these rules:
```
{
    position: absolute;
    width: 30px;
    height: 30px;
    background: lightgreen;
    left: 2px;
    mix-blend-mode: darken;
    pointer-events: none;
}
```

|  |
| --- |
| Colorizing the icon with mix-blend-mode |

Though there could be other methods too, like creating different colorized icons as their own images, or even having a spritesheet of all icons for the TardPad which could use background positioning to select the sprite ([see this CodePen for an example](https://codepen.io/chriscoyier/pen/GRpXLbd) which is the same technique, just used as an animation)

### Tooltip Grouping and Positioning
The tooltips for the two buttons could be grouped which means that they could not only be styled together, but there would be no ghosting between them when hovering from one button to the next:

|  |
| --- |
| The previous tooltip is visibly fading when moving to the next, creating a ghosting effect |

Another thing is that the tooltips are displayed as `top`. Given their placement so closely to the right of the UI, their messages actually fall outside of the UI, which could be a problem for any displays that fully contain the UI. Instead, `top end` should be used which will keep the tooltip entirely within the boundaries of the interface which I'll show in the gif below

The Tooltip class has general positioning for `top`, `left`, `right`, and `bottom` which will position the tooltip in those directions, centered relative to the target element. But with `start` or `end` modifiers such as `top end` or `bottom start`, these will align the start or end of the tooltips with the edges of the target instead of centering them

So instead of defining the buttons separately like this:
```html
<div class="right-icons">
    <div id="gameSettingsButtonContainer">
        <button
            id="gameSettingsButton"
            onclick="GameSettingsButton.toggle()"
            data-tooltipPosition="top"
            data-hideAfterInteraction="true"
            data-tooltipHtml='
                <div class="genericTooltip">
                    Game Settings
                </div>
            '
        ></button>
    </div>
    <div id="controllerStatusButtonContainer">
        <button
            id="controllerStatusButton"
            onclick="showControllerStatusModal()"
            data-tooltipPosition="top"
            data-hideAfterInteraction="true"
            data-tooltipHtml='
                <div class="genericTooltip">
                    <span id="controllerTooltipText">TardPad Status</span>
                </div>
            '
        ></button>
    </div>
</div>
```

You could define them like this. Note that only the `data-tooltipPosition` and `data-hideAfterInteraction` attributes have been removed, and `data-tooltipGroupId` attributes have been added:
```html
<div class="right-icons">
    <div id="gameSettingsButtonContainer">
        <button
            id="gameSettingsButton"
            onclick="GameSettingsButton.toggle()"
            data-tooltipGroupId="bottomRightIcons"
            data-tooltipHtml='
                <div class="genericTooltip">
                    Game Settings
                </div>
            '
        ></button>
    </div>
    <div id="controllerStatusButtonContainer">
        <button
            id="controllerStatusButton"
            onclick="showControllerStatusModal()"
            data-tooltipGroupId="bottomRightIcons"
            data-tooltipHtml='
                <div class="genericTooltip">
                    <span id="controllerTooltipText">TardPad Status</span>
                </div>
            '
        ></button>
    </div>
</div>
```

Then set the styles for `bottomRightIcons` in the Tooltip's `groupSettings` at the top of the script:
```javascript
Tooltip.groupSettings = {
    playerInput: {
        delayMs: 800,
        hideAfterInteraction: true,
        position: "left",
    },
    // Settings for the new group
    bottomRightIcons: {
        hideAfterInteraction: true,
        position: "top end",
    },
};
```

This would remove the ghosting and keep the tooltips pointing in towards the rest of the interface:

|  |
| --- |
| The tooltips working together with the new settings |

src/game.html
```diff
@@ -10054,18 +10064,36 @@
...
+                    merchant.consecutiveFloorsAbsent = 0;
+                } else {
+                    merchant.isActiveOnFloor = Math.random() < 0.35; // Chances at spawning normally 
+                    
+                    if (merchant.isActiveOnFloor) {
+                        merchant.consecutiveFloorsAbsent = 0;
+                    } else {
+                        merchant.consecutiveFloorsAbsent++;
+                    }
+                }
```
packardbell95, line 10084: This same logic could be handled in half as many lines:
```suggestion
                // Always spawn on the first floor, or after 3 floors of absence
                if (floor === 1 || merchant.consecutiveFloorsAbsent >= 3) {
                    merchant.isActiveOnFloor = true;
                    merchant.consecutiveFloorsAbsent = 0;
                } else {
                    merchant.isActiveOnFloor = Math.random() < 0.35;
                    merchant.consecutiveFloorsAbsent +=
                        merchant.isActiveOnFloor;
                }
```

src/game.html
```diff
@@ -105,6 +105,10 @@
     <div id="screenWipe"></div>
 
 <!-- In-game -->
+    <div id="vampireFocalPoint"></div>
+    <div id="vampireFocalPointDesaturationOverlay"></div>
+    <!-- Note to packardbell95: The vampire focal point stuff was moved up here because after I made changes to the bottombar, the colors for the animation broke -->
```
packardbell95, line 110: Noted. Thanks for calling this out :v:

src/game.html
```diff
@@ -10054,18 +10066,28 @@
...
-                (floor === 1 || Math.random() < 0.35);
+            if (merchant.isAlive) {
+                // Always spawn on the first floor, or after 3 floors of absence
+                if (floor === 1 || merchant.consecutiveFloorsAbsent >= 3) {
+                    merchant.isActiveOnFloor = true;
+                    merchant.consecutiveFloorsAbsent = 0;
+                } else {
+                    merchant.isActiveOnFloor = Math.random() < 0.35;
+                    merchant.consecutiveFloorsAbsent +=
+                        ! merchant.isActiveOnFloor;
```
packardbell95, line 10077: One of my suggestions was actually bugged, so I fixed it here. The previous version wasn't incrementing the `consecutiveFloorsAbsent` properly, so the merchant would still go for long periods without spawning. I tested this by descending 1100 floors and checking the number of times the merchant was spawned, and he was never absent for more than 3 floors at a time

src/scripts/gamepad.js
```diff
@@ -630,7 +629,55 @@
...
+// Display connection status modal popup when clicking on the Tardpad icon
+function showControllerStatusModal() {
+  const pads = getPads();
+  const isConnected = pads.length > 0 && pads[0];
+  
+  let content;
+  if (isConnected) {
+    const gp = pads[0];
+    content = `
+      <h3>🎮 Gamepad Connected</h3>
```
packardbell95, line 642: Excuse me, but is that an _emoji_ in this _retro text mode game?!?!_

packardbell95 (approved):
Changes look great

I tested the merchant and can confirm that he will spawn after three levels of absence

I also tested the gamepad:
- Clicking on the icon when it's red displays the "No Gamepad Detected" modal
- Plugging in a gamepad turns the icon from red to green
- The game is playable after plugging the controller in
- Clicking on the green icon shows a modal with the gamepad's ID
- Removing the gamepad dongle turns the icon back to red

Nice work :+1:

PR #159, opened 2025-09-21
Embellishing title screen, adding simple animation when descending floors

packardbell95:
This PR includes two new updates:
- A new title screen
- A short animation when descending floors

No screenshots because that would spoil the fun :stuck_out_tongue_winking_eye:

This update also removes all of the title screen assets and logic from `game.html` and places it into its own data script. The title screen is unloaded once the game starts


TheMilkMan6669 (changes requested):
This is absolutely insane, wow. Great job here. One thing though... when quickstart is enabled, the title screen logic breaks and keeps the background animation overlayed on top of the game. So you can control the game, but you can't see anything. At first I thought this was a Chromium issue because it started like this when I launched it in Brave, but testing it in Waterfox and enabling quickstart reproduced the issue.

TheMilkMan6669 (approved):
Working great now. I don't think any further changes are needed.

PR #166, opened 2025-09-26
Menu and Inventory Updates

packardbell95:
This PR makes a few modifications to the menu system. These include the following:

#### Removing breadcrumb hard-coding from menu titles
Menu titles had hard-coded breadcrumbs for their titles. The `Menu` class now provides these based on the active menus natively by calling `Menu.getFullTitle()`, so we don't need to worry about pairing the titles with how menus are opened moving forward

|  |
| --- |
| Breadcrumb titles when the menu is open |

#### Showing equipped details in tooltips, colorizing the Close button
This PR includes changes that have the sidebar inventory's tooltips call out if an item is equipped or not, including on which hand a ring has been equipped. The Close button is now a muted red instead of black by default to help indicate that the player is looking at a submenu

|  |
| --- |
| Selecting gear now shows "Equipped" messaging in the tooltip |

#### Hiding the scroll buttons in the inventory sidebar by default
When starting the game with Auto Diceroll enabled, the Scroll Up button would be displayed in the sidebar when it shouldn't appear at all

|  |
| --- |
| The up arrow is shown when starting a new game with Auto Diceroll enabled |

#### Separating the menu cursor from the displayed text
The menu had the cursor displayed along with a given option's text. Any styles applied to a menu option would also be applied to the cursor. This meant that when an item was unavailable, the cursor would also be crossed out and muted. This PR separates these two so the cursor can always be seen

|  |  |
| --- | --- |
| **Before:** The cursor is crossed out when the selection is disabled | **After:** The cursor is always clearly shown |

### DIFF
--- a/src/scripts/Menu.js
+++ b/src/scripts/Menu.js
@@ -326,7 +326,6 @@ class Menu {
             const isSelectedLine = index === this.#selectionIndex;
             const classes = [
                 "option",
-                ...(option.className ? [option.className] : []),
                 ...(isSelectedLine ? ["selected"] : []),
             ];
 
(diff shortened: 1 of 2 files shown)


TheMilkMan6669 (approved):
Changes all look good to me, the new breadcrumb system in place is very convenient. Also  much appreciate the equipment status callouts in the sidebar tooltips as well.

PR #167, opened 2025-09-28
New torch anim, Cup of Lean FX, ambient SFX, etc

TheMilkMan6669:
This PR includes a few nice changes, such as:
- New torch animation to replace the old placeholder one
- SFX/VFX added for when you drink a Cup of Lean - filter lasts for 60s (or 60000ms)
- Replaced gamepad emoji in the TardPad status modal with the regular icon
- Random ambient SFX will play between random time intervals (entries are under the `const sfxAmb` object)
- Nerfed chances at running away from battle somewhat
- Restructured audio folder + updated preloader entries accordingly

**_!! FIXED !!_** There is, however, a bug where if you enter battle with the lean effect on prematurely the environment won't darken until you you are hit by the enemy. I'm sure @packardbell95 would know how to go about fixing this since he handled the darkening effects when creating the pseudo-3D renderer.


VocaPepper (approved):
I corrected some incorrect file paths in the preloader, but everything else looks good. Can't wait to buy a bunch of lean for a drunk run.

PR #171, opened 2025-09-30
Improved Gamepad Activity Monitoring for UI Swapping

### LINKED ISSUE #132: [TardPad] Icons stealing priority when a gamepad is connected while using keyboard
Right now, if a gamepad is connected the UI will constantly be overridden when using a keyboard.

> Not ideal
> Image

This is most likely due to the changes made to the polling loop from PR #120 

I don't think this will be a difficult fix and will be looking into it soon.

VocaPepper:
This PR fixes #132 by improving how gamepad activity is tracked.

Previously a connected but idle controller could cause UI hints to revert to controller. Now only real activity (button/axis change) will switch hints.

### DIFF
--- a/src/scripts/gamepad.js
+++ b/src/scripts/gamepad.js
@@ -13,6 +13,7 @@
   let lastMenuDpad = { up: 0, down: 0, left: 0, right: 0 };
   let partyTalkLocked = false;
   let talkHoldStart = 0;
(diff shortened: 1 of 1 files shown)


VocaPepper:
Out of the three of us, I believe @TheMilkMan6669 has the most controllers for testing this change. I only tested with my Dualsense 5 and VKBs.

Issue #168, opened 2025-09-29
Spread out ambient SFX


TheMilkMan6669:
Ambient SFX tend to play a little too often. We should increase the minimum delay in which a sound will play, and probably the max delay as well. The next sound also tends to play immediately after a battle ends, and I suspect it might be because the next sound is being queued in the background or something like that.

Issue #170, opened 2025-09-29
Add distance-based audio cues for NPCs


TheMilkMan6669:
It would be a nice detail to have audio cues play when there's an NPC nearby. Like for example, if there's a Gambler nearby, you might hear a diceroll in the background. This would add to the immersion, as well as giving the player a way to tell if there's an NPC on the floor without needing to use a torch.

The Gambler in particular might be a tricky one to handle since he technically sorta exists on every floor that his spawn chances give him, just set to hidden if you're missing the 200btc requirement. We'll have to see how this plays out in practice.

Issue #169, opened 2025-09-29
Prevent ambient SFX from playing on title screen + NPC interactions


TheMilkMan6669:
Ambient SFX will play on the title screen and during NPC interactions. The threat must be eliminated.

PR #172, opened 2025-10-03
Rebalancing, random encounter grace period, NPC audio cues, misc

TheMilkMan6669:
This PR does the following:
- Rebalancing of weapon/armor stats/requirements. In my testing, floor 1 serves as a not-as-difficult character setup period, then things get challenging from floor 2 onwards. This will still need more playtesting.
- Pricing adjustments for items, weapons, and armor.
- A grace period/safeguard for random encounters; after a battle, you will be able to take 2 steps without any random encounters, then the game will start rolling encounter chances again.
- Reworked ambient SFX scheduling so that it doesn't run the delay timer in battle or in menus. Now you won't often hear an ambient sound play right as a battle ends.
- Ambient SFX disabled during title screen and menu interactions.
- Ambient SFX delay intervals are more spread out. The next sound will play between 40-60s apart.
- Proximity based audio cues have been added for the Merchant and Erok NPCs. If you are 8 steps away from one of these NPCs, you will hear a reverberated sound effect play, telling you they are nearby.
- Once again fixed the softlock that occurs when you have auto diceroll enabled, a gamepad connected, and level up animations after levelling up from a battle.

This PR will close issues #168, #169, #170


packardbell95 (approved):
- ✅ Rebalanced stats and items feel less steep than before
- ✅ Items are more affordable, making merchant encounters more helpful
- ✅ Random encounters now have a grace range of two steps before they are re-enabled, which makes traversing the dungeon feel a lot less like enduring [The Paddling of The Swollen Ass](https://www.youtube.com/watch?v=IFEJvIfMFfQ)
- ✅ Ambient sounds
  - no longer appear on the title screen or in the menus
  - are more spaced-out and feel more natural
- ✅ The merchant and Erok will make a sound if the player is within 8 steps of reaching them
- ✅ Having auto diceroll enabled, a gamepad connected, and level up animations disabled after leveling up from a battle no longer softlocks the game

Overall, the changes are good

I also like how the sounds are arranged so that if both the merchant and Erok are within range at the same time, their ambient sounds don't interrupt one another

#### Random Ideas
A couple of ideas that came up while testing this that don't have to be addressed here, but I wanted to throw them out there:

##### Should the player's steps reset after descending to the next floor?
This is probably a moot point given that the 2-step range is already so low, but resetting would guarantee that the player would at least make it a couple of steps when on a new floor before getting into a fight

##### Thoughts on proximity-based NPC sounds
Currently, the proximity sounds occur as soon as the player is within range, but don't happen again until the player is out of range. This means that the player can land 8 steps away from an NPC, then step back and forth 9 steps away, then 8 again, to keep triggering the NPC sound. The sounds don't play if the player gets closer to the NPC either, which means that the sound can't be used as a location cue since it doesn't inform the player what direction the NPC is in

Rather, at some point, we could probably schedule proximity-based sounds once the player is within range, and keep playing them after a long interval. The sounds could be played louder the closer the player is to the NPC

With this kind of dynamic mixing, it would also be helpful to have a version of each NPC sound that is reverberated so that, as the player is further away, the reverberated version is played louder than the dry version, and the dry version could be louder once the player is close. This would help sell the atmospheric quality of the sound and feel more like the player is in a labyrinth

---

Overall, nice work :+1:

TheMilkMan6669:
@packardbell95 I've implemented some of those suggestions you brought up. You can check the recent commits, but most notably, NPCs now have proximity based volume scaling for their audio cues and they repeat on a time interval between 10-20 seconds.

packardbell95 (approved):
Trying out the latest changes, the NPC sounds now work based on proximity. Moving closer towards either the merchant or Erok will make their sounds louder and moving away will make them quieter within an 8-cell radius. Very cool 😎

PR #175, opened 2025-10-14
Boulding Ball trap

TheMilkMan6669:
The TardSpire has been met with a terrible new threat...

## THE BOULDING BALL (dun dun dunnn)
Floors now have a chance to spawn an Indiana Jones boulder type object, with the twist being that it's not just any giant boulder; this one is a giant bowling ball. Feature list below.

- A single Boulding Ball will spawn in any hallway (horizontal or vertical) that is 4 or more empty spaces long, with a 60% chance of spawning on each floor
- A single wall gap will be generated to the side so that the player has a way to move around it without tanking a cheap hit
- The player will lose equivalent to one third of their max HP once hit, ensuring this will always remain a constant threat no matter your stats
- Proximity-based SFX will play, with the volume dynamically adjusting depending on the distance away from the player. This will be important to rely on when the player is not using a torch.
- Unlike the Vampire, the Boulding Ball rolls in realtime, finally making SPD an important stat to maintain
- The Boulding Ball will check for entities in its path (chests, healtiles, etc), ending its path once it's made contact without destroying anything. If that entity is removed from the map (i.e., after opening a chest), the Boulding Ball will be able to roll past that point. This could potentially open up some strategic dungeon navigation situations.
- The menace is impervious to all damage from plastic explosives.
- The Boulding Ball will cease to move while in battle, a menu, a level up animation, or a game over state.
- The minimap ASCII character is animated, but is misaligned in the tooltip. I will need @packardbell95 to help sort this one out. (edit: bug has been fixed)
- When hit by the Boulding Ball, a unique graphic will slide into the upper right area of the viewport.

https://github.com/user-attachments/assets/4a66b75b-40b9-4274-b4cd-40a856b270d5


src/styles.css
```diff
@@ -1430,6 +1454,53 @@ a:link {
...
+    50.01%, 75% {
+        content: "o";
+        transform: scaleY(1);
+    }
+
+    75.01%, 100% {
+        content: "o";
+        transform: scaleY(-1);
+    }
+}
```
packardbell95: Since the `content` never changes, these could technically be placed [within the pseudoelement directly](https://github.com/packardbell95/tardquest/pull/175/files#diff-4753b2e6427841425517139f09f1342c320338443447af46169dc0f43a0f2cf3R1461) rather than included in each keyframe. Though this is minimal and won't change any current behavior

The animation itself also flips the "o" upside down twice with an equal duration (0-50%, then 50.01-100%). This could be reduced with a jumping animation too:



```suggestion
        content: "o";
        display: inline-block;
        animation: boulderRoll 0.3s steps(2, end) infinite;
        color: #8B5A28;
    }
}

.mapCellDetails .enlargedTile.boulder::before {
    top: 0.4ch !important;
}

#minimap .boulder {
    width: 1ch;
}

@keyframes boulderRoll {
    0%, 50% {
        transform: scaleY(1);
    }

    50.01%, 100% {
        transform: scaleY(-1);
    }
}
```

(I think GitHub's multi-line diff is broken, so this might not show up as nicely as it should)

packardbell95 (approved):
Changes look good. [I did remove the Sound Forge file that was included](https://github.com/packardbell95/tardquest/pull/175/commits/256a813119874b20f611ce16d19fd540e88449b7) since nothing was referencing it and just appeared to be some RIFF WAVE header anyways. That note aside, this addition is excellent. Nice work! 🙌

VocaPepper (approved):
I love more dumb ways to die!

PR #178, opened 2025-10-21
Free-roaming enemies, wait button, etc

### LINKED ISSUE #177: Fix chest item drop chance system
The intent with treasure chests is that items have a percent chance at dropping once a chest is open. These rates aren't respected though. Even if an item is set to drop at a 0% rate, you can still receive that item from a chest. 

This is especially a problem with the Carrier Pigeon, because the game doesn't check your connection to the Pigeon API once picked up from a chest. If you use a chest dropped Carrier Pigeon, it will pop an error in the battle log and result in the persuasion mechanic bring completely busted for the rest of the playthrough. Personally I want the Pigeon to have a 0% chest chance anyways, but you guys are welcome to make some API check updates regarding chest drops for the Pigeon anyways if you want.

TheMilkMan6669:
Free-roaming enemies, wait button, etc
Decent sized update here. List of changes:

- No more random encounters; enemies now free-roam the map, of course with sceneArt included. They all share the same generic sceneArt though, don't get too excited. 
- Enemies will actively pursue and follow you if you get near.
- The Ring of Stinky has been reworked to cause enemies to have an 80% chance of avoiding you per step, rather than pursuing you. It is still easy to corner them, so you're not ganking yourself out of levelling up entirely.
- When a few enemies have been killed, a random number of them will respawn (allegedly).
- Added a wait button, which will allow enemies to pass by if you so wish. This will also count towards the vampire step counter. Spacebar on keyboard, B on Xbox, Circle on Dualsense.
- Fixed and reworked the chest drop system for items. There is a simple "chestDrop: True/False" flag now, which you can define under the item/ring entries in the inventory object. I felt that percent chances for individual item drops was redundant and made things overly complicated anyways. Fixes #177 
- It is now much simpler to adjust chest loot type (BTC, item, ring) percent values.

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -12284,19 +12284,19 @@
                         playerAttackCooldown = false; // Reset for level up screen
                         player.leaveCombat();
                         setTorchOverlayVisibility();
-                        startLevelUpAllocation(() => npcMove());
+                        startLevelUpAllocation(() => render());
                     } else {
                         player.leaveCombat();
                         setTorchOverlayVisibility();
                         if (battleTransitionAnimationsEnabled) {
                             playerAttackCooldown = false;
                             animBattleTransitionStart(() => {
-                                npcMove();
+                                render();
                             }, false, true);
                         } else {
                             playSFX('battleEnd');
                             playerAttackCooldown = false; // Reset when battle fully ends
-                            npcMove();
+                            render();
                         }
                     }
                 });


src/styles.css
```diff
@@ -2532,9 +2537,12 @@ a:link {
 .friendly {
     color: #61C9F6;
 }
-.enemy {
+.enemy { /* Unused class now, as of free-roaming enemies update? Not sure. -TheMilkman*/
```
packardbell95: This class is still used pretty heavily in the battle log

src/game.html
```diff
@@ -10116,6 +10426,209 @@
...
+
+            // Check if a position is in a boulder's path
+            isInBoulderPath: function(x, y) {
+                if (!boulder.isActiveOnFloor) return false;
+                
+                if (boulder.direction === 'horizontal') {
+                    return y === boulder.y && x >= boulder.hallwayBounds.start.x && x <= boulder.hallwayBounds.end.x;
+                } else {
+                    return x === boulder.x && y >= boulder.hallwayBounds.start.y && y <= boulder.hallwayBounds.end.y;
+                }
```
packardbell95: This could be shortened up a bit with a ternary if desired

```suggestion
                return boulder.direction === 'horizontal'
                    ? boulder.y && x >= boulder.hallwayBounds.start.x && x <= boulder.hallwayBounds.end.x
                    : boulder.x && y >= boulder.hallwayBounds.start.y && y <= boulder.hallwayBounds.end.y;
```

src/game.html
```diff
@@ -10531,6 +11044,17 @@
                     playSFX('scream');
                 },
             },
+            roamingEnemy: {
+                displayName: "Monster",
+                mapCharacter: "☠︎︎",
```
packardbell95: I thought this was an asterisk in-game until I saw it enlarged on the minimap

TheMilkMan6669: I couldn't really find a better one to copypaste honestly. It works good enough for me, but if you have any suggestions, go ahead and shoot

src/game.html
```diff
@@ -10116,6 +10426,209 @@
...
+                        const cell = MAP.getCell(newX, newY);
+                        if (cell && !cell.isSolid && (cell.type === 'floor' || (newX === player.x && newY === player.y))) {
+                            // Check collision conditions (excluding player position now)
+                            if (roamingEnemies.isAt(newX, newY)) return;
+                            if (merchant.isAt(newX, newY)) return;
+                            if (gambler.isAt(newX, newY)) return;
+                            if (erok.isAt(newX, newY)) return;
+                            if (vampire.isAt(newX, newY)) return;
+                            if (pigeon.isAt(newX, newY)) return;
+                            if (boulder.isAt(newX, newY)) return;
```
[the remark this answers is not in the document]
TheMilkMan6669: We'll worry about that in a future update I suppose

TheMilkMan6669:
>We could hold off on introducing the Vampire until later in the game, and simply cap regular enemy respawns after some set amount of kills early on and/or severely slow down the respawn rate after a kill threshold has been reached. This might make the game a little easier to approach for newer players, allowing them more time to explore the map

Sounds like a good idea to try out in the future. We can make this a separate update.

> Tie Map Movement Behavior to Specific Enemies
The enemies have some real personality. This should be reflected in their movements on the map. For instance, the Snail Sentinel could move slowly, if it even moves on a turn at all. Meanwhile, the Wang Rat or Stupid Dog could get extra range per turn (eg: up to two spaces instead of one) since they are bigger and more agile creatures

This could also tie into how far away the player is before the enemies notice them for a chase (if they even want to chase the player)

This is a fun idea, I like it. To add onto this, it might be a good idea to recolor the empty tiles on the minimap to yellow or something to represent their FOV, like how you see vision cones in other games. This could help introduce more of a stealth oriented playstyle, as non-viable as it may be.

>Introduce a Monster Journal-Type Item (Pokédex, sorta)
...

Nice to see this come back into conversation, it's been in the back of my mind for a while. Once we throw (at least a basic form of) status effects into the game, that would be a perfect time to introduce something like the TarDex. Or whatever we end up dubbing this Pokedex clone mechanic. All of these are good ideas.

>Regarding identifying known monsters on the minimap, maybe revealing the map with a torch would show where monsters are and keep them ambiguous, but would only allow the player to identify them if they are within the player's line of sight. Otherwise, the torch might be too magical

I like this

>Maybe a crystal ball type of item could reveal monster positions and show their types, but not reveal the entire map as well

I like this as well. Needed an excuse to record a wizard voice for this game anyway

packardbell95 (approved):
Changes look good. Trying out the latest after my last review:
- Enemy portraits no longer show up before the curtain drop
- There's a gap between the directional arrows and the Wait button in the UI to minimize accidental clicks
- The Wait button is more legible than it was previously
- Roaming enemies, including the vampire, do not move after a battle has ended (they will move after the player has successfully ran from a fight, but this feels like a normal outcome)

Excellent work 👍

PR #179, opened 2025-10-21
Various Minor Freeroam Enemy Updates

packardbell95:
This PR makes the following updates:
- Roaming enemies are now tracked within an object instead of an array to avoid conflating index position with ID which would cause some battles to not even start
  - Enemies start battles based on ID rather than map position
  - Removing roaming enemies now happens with `delete` rather than `.splice()`
- Fixed bug where enemy kills were not always being counted
- Simplified some logic, particularly around respawning rules
- Enemies now spawn a minimum of two spaces away from the entrance after exiting to a new floor to avoid starting battles upon descent

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -10427,16 +10427,18 @@
         ];
 
         const roamingEnemies = {
-            enemies: [],
+            enemies: {},
             maxEnemies: 6,
             nextId: 0,
             enemiesKilledOnFloor: 0,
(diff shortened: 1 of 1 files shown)


TheMilkMan6669 (approved):
Seems to be working good on my end. Good catch on preventing enemies from spawning on top of you when entering a new floor.

PR #187, opened 2025-11-18
Kill counter changes + potential control display overflow bugfix

### LINKED ISSUE #173: Control Details Overflow When Gamepad is First Used
Noticed this while testing https://github.com/packardbell95/tardquest/pull/171, though that PR is not the cause:

When the UI is first interacted with, the controls listed in the bottom left corner of the UI will briefly display two sets of controls which cause the list to extend below the interface

|  |
| --- |
| After touching the gamepad for the first time, the control section briefly overflows below the bottom left corner of the UI |

This is not a critical bug because most people are playing the game in an iframe which cuts off at the bottom of the UI normally

The issue is likely caused by one set of controls being displayed before the other set is hidden. Some potential fixes include:
- Swapping the order so that no controls are displayed briefly instead of both
- Absolutely positioning the controls within the controls section so that if two are displayed at the same time, they stack on top of each other instead of causing an overflow

### LINKED ISSUE #185: Boulder Kills Count Toward Player Kill Statistics
### Description
When the boulder kills enemies, those kills are incorrectly being added to the player's total kill count. The boulder should deal damage independently without affecting player statistics.

### Expected Behavior
- Boulder kills should NOT increment the player's kill counter
- Only enemies killed directly by the player should count toward their statistics

### Actual Behavior
- Enemy killed by boulder → player kill count increases
- This inflates the player's actual combat performance metrics

### Steps to Reproduce
1. Start a new game
2. Navigate to a location where the boulder is active
3. Allow the boulder to collide with and kill an enemy
4. Observe the player's kill count under minimap
5. **Expected**: Kill count unchanged
6. **Actual**: Kill count increments

### Root Cause
The boulder's collision/damage logic calls `roamingEnemies.killEnemy()` without distinguishing between boulder kills and player kills. The `killEnemy()` function increments `roamingEnemies.enemiesKilledOnFloor` regardless of the damage source.

TheMilkMan6669:
Simple list of changes:

- Kill counter has been moved to the minimap header. Closes #185 
- "Dungeon Floor" is now anchored to the left, with the kill counter to the right
- "Enemies Killed: (insert body count here)" is now simply represented by a unicode skull icon with the number beside it
- Minimap footer removed, class included
- Boulding Ball no longer counts towards the kill counter
- The kill counter will now persist across floors instead of resetting on each floor
- All references to `enemiesKilledOnFloor();` wiped off the face of the planet
- Potential fix for the control details overflowing when activating a controller for the first time; _**I was not able to replicate this issue myself, but I made an attempt here anyway by giving it absolute positioning. Please do let me know if this helped.**_ Closes #173

Screenshot 2025-11-17 230803

### DIFF
--- a/src/game.html
+++ b/src/game.html
@@ -23,7 +23,16 @@
         <div id="topLeft">
             <pre id="portraitArtOverlay" class="portrait-art hidden"></pre>
             <div id="minimapContainer">
-                <div id="minimapHeader" class="ui-header space-around">Dungeon Floor --</div>
+                <div id="minimapHeader" class="ui-header space-between">
+                    <span id="dungeonFloor">--</span>
+                    <span
+                        id="killCounter"
+                        data-tooltipposition="top"
+                        data-tooltiphtml='
+                            <div class="genericTooltip">Total Kills</div>
+                        '
+                    >0</span>
+                </div>
                 <div id="minimap"></div>
             </div>
         </div>
@@ -9924,6 +9933,7 @@
                 }
 
                 playSFX('scream');
+                updateKillCounter();
                 pigeon.say(
                     'I WAS ONLY THE MESSENGERRRRRrrrrrrrr... *fucking dies*',
                     true
@@ -10588,8 +10598,7 @@
 
             killEnemy: function(enemyId) {
                 roamingEnemies.removeEnemy(enemyId);
-                roamingEnemies.enemiesKilledTotal++;
-                updateMinimapFooter();
+                updateKillCounter();
             },
 
             isAt: function(x, y) {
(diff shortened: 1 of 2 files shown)


packardbell95 (commented):
Changes look good for the most part, though there are a couple of minor things I wanted to address. Generally speaking, the header updates could be simplified a bit. Rather than write the changes up here, I opened this [follow-up PR that's pointed at this one](https://github.com/packardbell95/tardquest/pull/189)

Also, I noticed that the controller overflow issue still seems to be present. We can tackle that elsewhere though since there might be better ways for us to organize the information in that panel which would be a bigger update anyways:
|  |
| --- |
| Controls still overflowing when a controller is plugged in |

PR #198, opened 2026-03-21
Alpha State of Core Engine Updates for TardQuest

packardbell95:
This draft is introducing a number of changes to the core engine of TardQuest. The largest items that are introduced include:

### New Battle System
|  |
| --- |
| Throwing down against a duo of Wang Rats! |

The battle system now supports many-on-many fights. Enemies can have up to 6 members in their party, and they want to throw down

- Enemies just hurl punches at the player, though they're technically capable of other moves
- The player can...
  - ...attack the enemy (also try this with the Magic Pencil equipped)
  - ...use battle-relevant items
  - ...persuade enemies (or at least the player can. There is currently no text input prompt, but that will be coming back in a future update)
  - ...throw a smoke bomb to run away (currently works 100% of the time due to debug code, will change soon)

When running from enemies, they will become stunned for a few turns
|  |
| --- |
| Stunning an enemy lets you get away, and reveals more info about the enemy in the minimap |

Members take turns fighting with faster members having turns sooner. When two members have equal speed, a randomly-generated tiebreaker is employed

Currently, queued moves cannot be overwritten in the UI, but will in a future update. Similarly, specific enemies cannot be targeted for attack, and party members can't be selected for item use, but these will also change

Enemies do not yet have death animations or visible health bars over their sprites

### New Entity System
Previously, TardQuest stored entity data in discrete objects. This update introduces a new system: the map is just made of walls and floor. Objects, creatures, and NPCs that appear on the map now exist on it and are known as `MapEntities`

A `MapEntity` is an object with a set of properties and functions that dictate how it exists and behaves. For instance, a `MapEntity` with an `.onTouch()` handler can be touched (such as with enemies and NPCs), or `.onEnter()` can be stepped on (such as with Healing Tiles)

`MapEntities` also can have different movement rules attached. While a few are present in the engine, the current one is a chase routine where, if an enemy spots the player, it will chase the player until they fight. This will be tweaked to allow for alternate behaviors and backoffs in the future

|  |
| --- |
| Pathfinding test showing entities walking in patrol routes, avoiding each other's paths |

A `MapEntity` will commonly have a `party` property. This is the set of party members that make up the entity. Party members are the teammates that you fight with in battle. The first party member, which is referred to as the party's `leader`, is the member whose details dictate how the whole `MapEntity` appears and operates. For instance, if a `MapEntity` has a Wang Rat as the `leader`, the whole entity will appear as a Wang Rat to the player. Similarly, with the player's map entity (`playerEntity`), the `leader` is a `player` party member. This `player`'s attributes control things in the game, such as the viewing depth that's used for the viewport renderer. This means that any status effect or equippable object can potentially change the operation of the game itself

It's best to think of it like: the `leader` is the driver and the `MapEntity` is the car that they are driving. The car represents the driver, even if others are along for the ride

### Stat Organization
Speaking of party member attributes, party members now have their stats organized like this:
- Stats (player's variable properties)
  - Progression (attributes tracking the player's progress)
    - Level
    - Experience
    - Total Experience (Tracked just for the API)
  - Core (practical capacities)
    - HP
    - Max HP
    - Defense
    - Strength
    - Persuasion
    - Endurance
    - Speed
    - Luck
- Traits (fixed values related to a party member)
  - Sight Range (how far forward a party member can see)
  - Sight Sensitivity (light sensitivity, unused)
  - Field of View (how far around themselves the entity can see)
  - Hearing Range (how far an entity can hear a sound, not yet used)
  - Persuasion Attempts (the number of attempts to use persuasion in battle, not yet implemented)
  - Hands (what equipable appendages the party member has, unused)
  - Can Attack (whether the party member can attack in a fight)
  - Can Persuade (whether the party member can persuade in a fight)
  - Can Run (whether the party member can run from a fight)
  - Can Use (whether the party member can use items in a fight)
  - Can Equip (whether the party member can equip gear in a fight, unused)

Because of the party updates, the inventory is handled differently too: each `MapEntity` has an `inventory` of items that everyone in the party holds. This includes `bitcoins`, `items`, `weapons`, `armor`, and `rings`. Each party member can equip gear which becomes an `equipped` property on that party member. This means that it gets removed from the `inventory.contents` while it's equipped by a member. Anything that requires counting inventory items therefore needs to also count whatever the party members are currently holding as well

Party members also now all have animated portraits which appear when they are speaking

|  |
| --- |
| A few of the enemy portraits that appear in this update |

### Engine vs Game Organization
This update is taking some steps towards splitting the responsibilities of the TardQuest system into the `engine` and `game`. The `engine` is the underlying workings of the game's mechanics where the `game` portion uses the engine's components to create the TardQuest experience

A good example of where to see this split is in how NPCs are created. The `MapEntityNpcFactory` is used to generate the game's NPCs and plug them into the map. This factory uses the engine's `MapEntityBuilder` to make sure that entities are being built with all required or expected parameters. This setup makes it simple to generate NPCs (or enemies and map features) that have complex underlying behavior

### Other Changes

Anyways, here is an incomplete list of other changes so far:

#### Testing
- Added more context renderers and helpers
- Errors are no longer printed inline, but are behind a button that reveals an Error Details modal, comparing the expected and actual values where applicable

#### Game
- The battle log now indents wrapped lines to make entries easier to discern
- Healing tile icons now glow on the minimap to match the glow in-game
- After running from a fight, the enemy's party is identified in the minimap tooltip when hovering the enemy
- Persuasion works only when there is one enemy remaining in the enemy's party
- After persuading an enemy to be an ally, they will get their own pseudorandomly selected color used in their talk messages and portraits to distinguish them from each other, and from the tomato-colored enemies

#### Technical
- Inventory registry definitions have been moved to `src/data/`
- Functional components have been moved to `src/scripts/`
- Syntax updates are being employed to limit lines of code to 80 characters and to make things like quotes and curly brace use more consistent

#### Viewport Overlay Updates
- Viewport overlays (curtain drops, rat consumption, healing tile effects, etc) have been converted from GIFs and CSS rules into WebMs
- WebMs are handled through a new `ViewportAnimation` object
- `ViewportAnimation` references VTT metadata files that communicate when events should fire. Listeners are named after the event's name, with the prefix "on" in camelCase, eg the event "attackPeak" will call `onAttackPeak()` if it's present

#### Progress Bar Element
- Now supports a left-aligned `label` (currently used for the enemy health bars in the minimap after running from an encounter
- Numeric min/max values are now displayed with a maximum of 1 place digit (eg: the value `20.413` would display as "20.4", but `31.000` would display as "31")

#### SamJs
- Added an optional onEnded() callback function to `.speak()`

### Persuasion Note
The persuasion mechanic is just a placeholder that takes no extra input and only has a chance of working if there is one enemy left during a battle. This will be replaced later


VocaPepper:
Is there anything I'll need to update on the TardAPI server, or that you would like to see added?
I am currently [working on/testing V4](https://vocapepper.com/programs/tardquest.account-mode) which adds support for the TQO database backend for accounts (with the "Guest" mode still intact) among other small updates.

packardbell95:
> Is there anything I'll need to update on the TardAPI server, or that you would like to see added? I am currently [working on/testing V4](https://vocapepper.com/programs/tardquest.account-mode) which adds support for the TQO database backend for accounts (with the "Guest" mode still intact) among other small updates.

Nothing that needs to be implemented yet. I'm aiming to have these changes in a state where they'll work with the existing API. This is mostly going to involve handling things like pigeons and tombstones as entities in the new entity system

Some potential future changes that Milkman and I have been talking about would impact the API, though. In particular, I passed him the idea of making TardQuest more into a roguelite where there's a story and a goal. Players would descend into the Spire, reach the bottom, fight back out, and have an optional post-game experience where they can keep playing afterwards

This means that there wouldn't be any perma-death. So instead of collecting a player's name at the end of a run, the API could issue a unique ID to the player at the start of the game, and the game would continuously report back to the API with progress, similar to how it works now. The leaderboard would then be dynamic where scores would keep increasing in realtime rather than collected at just an ending. The player could change their name in-game at any point in the game too which could be passed back to the API which would update the leaderboard

Other stats could be tied into the leaderboard too, such as highest number of party members captured, total number of kills, total number of player deaths, total play time, etc, which could open up the possibility of having leaderboard categories beyond just score

These are all early ideas, but I think we'll try planning and formalizing them once we get through this round of changes first. Feel free to raise any questions or thoughts you may have about any of this

TheMilkMan6669:
@packardbell95 Sometimes it can be slightly difficult to tell who is and isn't an enemy on the battle queue UI despite your party members' colors matching their portraits. Have you thought about putting green and red lines under the icons in the queue to help differentiate them? Something kinda like this:
example
 Or alternatively, there can be an icon in the upper right corner of their portraits (green smiley for the player party, red evil thing icon for enemies) so it doesn't just look like a health bar.

packardbell95:
> @packardbell95 Sometimes it can be slightly difficult to tell who is and isn't an enemy on the battle queue UI despite your party members' colors matching their portraits. Have you thought about putting green and red lines under the icons in the queue to help differentiate them? Something kinda like this: example Or alternatively, there can be an icon in the upper right corner of their portraits (green smiley for the player party, red evil thing icon for enemies) so it doesn't just look like a health bar.

I definitely agree that the current state is hard to tell who's who. I think the bar idea is on the right track. Maybe a full border around each portrait icon would work better here, possibly with different background colors too

The backgrounds currently turn blue or red depending on who is acting and who is being acted upon, but they could be conflated with enemy/friendly alignment, so that will likely change too

I've been considering some other things too, like inverting enemy portraits. But your idea of having a friend/enemy icon might be best. I was considering putting icons on these portraits to show what actions are currently queued for a given team member for the current move, but that might result in some visual overload. Plus, those details can be displayed in the stats sections, and within the tooltip that I'll be including for the battle queue as well

TheMilkMan6669:
To-Do list:

- Enemy attack animation(s)
- Party attack animation(s)
- Persuasion animation (I may have an idea)
- Enemy on-hit damage frame

Anything else to add here?

TheMilkMan6669:
Few things I want to point out after trying out the latest commit as of [cb415d8](https://github.com/packardbell95/tardquest/commit/cb415d80cd248df29999dfc132a4b83b0af8bdbc):

- The game will recognize USB devices other than game controllers and end up prioritizing my external speakers instead, forcing me to turn them off so the game would recognize my controller. This was an issue we were having a while back when we were first introducing gamepad support, which I believe @VocaPepper was able to mitigate iirc. You two can communicate here as needed. There also isn't a way to use the wait function on a gamepad here anymore either. Used to be bound to face button right, but now it just opens up the settings menu.

- The TardBoard score submission screen also doesn't display when I die, whether I'm using the live server plugin via VSCode or in the standalone Extra Tarded Edition build. Haven't tested when hosting on my website yet.

- Battle music doesn't play when fighting mimics. It seems to play fine everywhere else where applicable.

- Balancing still needs work. I think enemies need to either hit a bit less hard on floor 1, or have a bit less HP. One or the other. Maybe we should think about buffing flee chances too, especially now that we have group fights.

PR #215, opened 2026-08-17
Graphical Overhaul for the SceneRenderer and Portrait Systems

packardbell95:
<video src="https://github.com/user-attachments/assets/01795050-b263-4d04-86e6-92e5cd48b9e3"></video>

# Graphical Overhaul
To address the graphical limitations and performance problems that were caused by rendering ASCII art as HTML elements, the systems that have relied on these methods (namely the scene renderer and the portrait handler) have been replaced

A major goal of this PR was to remove ASCII-based rendering from the game entirely and replace it with rasterized image rendering instead. This avoids rendering within HTML itself which was the source of many performance problems and graphical inconsistencies. It was also a huge pain to edit graphics as text and CSS within a text editor. Instead, using images and sprite sheets, we will now be able to edit images faster and much more easily

Another goal was to extend the capabilities of the renderer itself without compromising the graphical aesthetic of the game. The previous renderer it was a pseudo-raycaster which would draw pre-made wall sections based on the player's position and facing direction within the map. This meant that there was no way to render, for instance, a half-turn or even a half-step, which hampered the sense of movement for the player, while also being a pain to simply add a new wall texture since that would have to be manually drawn at every supported distance and angle

With this update, not only do we have an actual raycaster at work, but it also supports things like textured surfaces and dynamic fog and lighting. This will allow us to create different kinds of environments, from dark and dank dungeons to fully-lit fluorescent office interiors and virtually everything in between

### Scene Renderer Overhaul
  - ASCII art has been removed from the game (except for the Gambler's dice throwing animation)
  - The scene renderer is now a proper raycaster instead of a pseudo-raycaster. This includes effects, such as wall darkening and Cup of Lean drunkenness
  - Rather than a glowing border around the frame, the player's torch now emits an actual light within the scene
  - The raycaster supports environmental fog and lighting, both statically and dynamically. In other words, a map's cell can have variable colored lighting and fog with variable intensity/density. Entities can also emit their own light, such as the healing tiles and sigils, or their own fog, such as the player when the Ring of Stinky is equipped
  - Walls, ceilings, and floors can be textured
  - Entities can display their own sprites as billboards or surfaces which can be composited together
  - Sprites can be animated and directional (8 degrees of rotation). This is currently used by roaming enemies and boulding balls
  - The player's sight sensitivity and range now factor into the rendered scene
  - The viewport background is also handled by the scene renderer rather than being a CSS-positioned background image
  - The scene renderer also supports level-of-detail dropoffs for sprites, so sprites can become more or less detailed based on their distance from the player as well as the player's visibility traits. This is currently only in place for roaming enemies, the exit arrow, gravestones, treasure chests, and the vampire bat
  - Battle animations are also now handled by the scene renderer rather than CSS. Animations will be further tweaked in a future update
  - The scene renderer supports variable rendering resolutions (eg: use `SceneRenderer.setStripSquare(1)` to make everything crisp and `SceneRenderer.setStripSquare(24)` to make the game look like it's on an Atari 2600)

### Minimap
  - **Performance update:** Reduced the number of updates that the minimap performs per-render so that it only updates the cells that have changed rather than every single cell on the map
  - **Bugfix:** For entities that move in realtime, the minimap now identifies their previous and current coordinates so that their cells get rerendered properly

### Art Updates
  - Since the renderer now supports angles, textures, and animations, several sprites have been updated. For instance, the bloody corpse that roaming enemies are reduced to when hit with an explosion are now a billboard skeleton and a pool of blood floor texture that are composited together

#### A Note about the ASCII-to-Image Conversion
Some of the conversion didn't quite go as well as I'd hoped, Particularly around images that used ASCII characters for shading. This is because these glyphs depict dithered shades, and rendering them at different sizes interfered with how these appear. So, in some cases, they have some unusual-looking patterns. However, some of these, specifically for the portraits, have been edited to have actual shaded pixels rather than dithering. I think over time, the ones with dithering will be modified to use actual shades instead since it looks the most appropriate. Plus, with how most of the art consists of white outlines and black fills, I think we may want to try different approaches like what was taken with the roaming enemy: keeping a white outline, but having a color fill for the body rather than solid black

But, since graphical styles will have more flexibility following this update, I think we can hold off on aesthetic decisions around these until later

### Movement
  - Player and enemy moves are now rendered in half-steps which gives a much stronger sense of movement. These also happen in tandem, so the player can see enemies actually walking past them
  - The player's default movement timeout has been raised from 100ms to 200ms. It felt about 200ms previously due to browser lag in the previous versions

### Portraits
  - All portraits are now based on sprite sheets as well and are drawn to their own canvas element
  - Each sheet consists of vertical frames that are 220x168
  - Animation and timing behavior is still defined within `TARDQUEST_PORTRAIT_ANIMATIONS`
  - Portrait animations are handled by a new `Portrait` object
    - Portraits can be shown with `Portrait.show()` to show a specific portrait or `Portrait.showPartyMember()` to show a specific party member's portrait
    - Party members have their own `portrait` object which contain the portrait's `name`, and may also include a `tint` color (used for enemies which are a tomato red and pseudo-randomized for your own party members) as well as a `flipped` parameter (used to show enemies facing to the left during battles)
  - Portraits also now have a text header that display the name of the party member or some other similar name/message (eg: "Game Over" during death)
  - Portrait animations can now hand off to another animation via `nextAnimation`. Specifically for petting Erok, this was handled with a `setTimeout()` previously, but now the portrait renderer knows that it should just start playing Erok's idle animation after petting has completed

### Misc
  - Timings for viewport overlay WebMs have been tightened to match predicted frame values
  - The player's icon will now appear in a half-rotation in the minimap when turning
  - Rings specifically have support for dynamic lighting and fog effects. This is currently used for the ringOfStinky which surrounds the player in a nasty fog. This functionality will be extended to other equipment at some point later
  - Render profiler code is still lingering in this update which has been used to make performance updates during development. This will be removed at some point later, probably after another round of performance updates
  - Fixed an `instanceof` miscalculation in the tooltip generator

## Things to Try
A new `/demomap` cheat code has been added to try out some demo regions. Try these:

| Cheat | Description | Preview |
| --- | --- | --- |
| `/demomap hod` | **The Hallway of Doom.** This was seen in a previous update where boudling balls roll back and forth, colliding into each other. The player has to make it through them to reach the exit. The new thing to see here is the 8-degree sprite rotation for the boulding balls which make them look somewhat more lively. Timing animations may not be quite aligned with the movements yet, but the effect is still an interesting one |  |
| `/demomap moss` | **The Mossy Crypt.** This is a randomly-generated map just like normal, but with randomly-placed ceiling tiles and different noise-generated surface textures to simulate moss and stone. This map also shows off wall torches that produce dynamic light so that you can make your way around in the dark. Try blowing up a wall with these torches attached and observe the light change around the destroyed area |  |
| `/demomap bb` | A fun undersea adventure! ;D This map demonstrates a map that's fully lit at the renderer's full resolution, plus some dynamic lights and a simple animation | ??? |


TheMilkMan6669:
This is insane. Love everything about this new renderer, I think you've struck a very nice balance between having a proper "3D" space and keeping the game's art style intact. The new sprite animations look sick and I love how you can see enemies physically sliding around. There's a lot of potential to be had here for feature implementations in the future as well. Also love the demo maps, bb especially is cool. I'm guessing you were testing out some stuff for the quest scripting we have planned.

I don't know the full extent of what this new engine is capable of, but it makes me think, maybe we could have a brief animation for when you open the inventory screen where the player looks down and opens up his trench coat, or looking down and inspecting your currently worn armor, or whatever other useless but cool interactions could be had.

### **_Encounterd bugs + general concerns_**

- If you open a chest, the icon doesn't disappear from the minimap until the player moves. Same deal with NPCs that are meant to disappear after being interacted with, like Erok and the Gambler. Functionally they disappear, just the minimap doesn't update immediately.
- When you die, you get the vampire battle portrait instead of the laughing game over death skull.
- If you fall down a cracked floor, the battle log will print the "you died on x floor" dialog many times.
Screenshot_20260816_165542

- When an enemy is hit during battle, their sprite is fully colored in and you don't get to see any of the detail until the last couple frames. Not sure if intended, but this could be adjusted.
Screenshot_20260816_165457

- If an enemy is blown up during battle, they won't leave behind a pool of blood. NPCs and Mimics (outside battle, during exploration) will.
- If you use C-4 during battle while an NPC is nearby, they will also die. Not sure if intentional, or even exclusive to this branch.
- Some world elements, like the vampire's spawn sigil, or ceiling textures, aren't dimmed out during battle. This can be visually confusing IMO.

packardbell95:
> This is insane. Love everything about this new renderer, I think you've struck a very nice balance between having a proper "3D" space and keeping the game's art style intact. The new sprite animations look sick and I love how you can see enemies physically sliding around. There's a lot of potential to be had here for feature implementations in the future as well.

Thank you! When I realized that I could implement the raycaster with a stepped and blocky look that matched the previous one's hallways, I had a really good feeling about it. I'm glad you're feeling the same

> Also love the demo maps, bb especially is cool. I'm guessing you were testing out some stuff for the quest scripting we have planned.

Hah, somewhat. I was mostly trying to see if tiled textures would look alright and didn't leave any distracting seams between walls or anything. I also used it to test standing textures that aren't actually a wall. So specifically, the doorway that Mr. Krabs is standing in has the string of flags hanging over it. That section is actually a whole wall texture and is the only spot in all three demos where there's a standing texture like this. It helped me find a graphical glitch where Mr. Krabs was actually being drawn on _top_ of the texture despite being behind it

I don't have any of the scripting stuff in this branch yet, but I do have a lot of ideas for it already. That's going to be one of the next things after this PR's closer to being settled

> I don't know the full extent of what this new engine is capable of, but it makes me think, maybe we could have a brief animation for when you open the inventory screen where the player looks down and opens up his trench coat, or looking down and inspecting your currently worn armor, or whatever other useless but cool interactions could be had.

That's an interesting idea. There might be some ways to do something like this. I'll give this some thought

> ### **_Encounterd bugs + general concerns_**
> 
>     * If you open a chest, the icon doesn't disappear from the minimap until the player moves. Same deal with NPCs that are meant to disappear after being interacted with, like Erok and the Gambler. Functionally they disappear, just the minimap doesn't update immediately.
> 
>     * When you die, you get the vampire battle portrait instead of the laughing game over death skull.
> 
>     * If you fall down a cracked floor, the battle log will print the "you died on x floor" dialog many times.
> 
> 
> Screenshot_20260816_165542
> 
>     * If you use C-4 during battle while an NPC is nearby, they will also die. Not sure if intentional, or even exclusive to this branch.
>
>     * Some world elements, like the vampire's spawn sigil, or ceiling textures, aren't dimmed out during battle. This can be visually confusing IMO.

Thanks for this list. Most of them I hadn't noticed, but they'll be pretty easy to fix

>     * If an enemy is blown up during battle, they won't leave behind a pool of blood. NPCs and Mimics (outside battle, during exploration) will.

Yeah, this definitely needs to be added. Should this happen if the battle is won with C4, or if any enemies are killed with it during the battle? I'm kind of leaning towards the former, although it could be funny in the rare case where someone could detonate an enemy's party member before fleeing battle. But maybe that could be a smaller pool of blood too. Maybe I'm overthinking it

>     * When an enemy is hit during battle, their sprite is fully colored in and you don't get to see any of the detail until the last couple frames. Not sure if intended, but this could be adjusted.
> 
> Screenshot_20260816_165457

Yeah, I definitely intend to adjust this. The battle transforms all need a little polish too. Like the scene renderer also handles skewing and rotation which is used with the Dance Club, but those don't look quite right yet either. But, it's close

TheMilkMan6669:
> Yeah, this definitely needs to be added. Should this happen if the battle is won with C4, or if any enemies are killed with it during the battle? I'm kind of leaning towards the former, although it could be funny in the rare case where someone could detonate an enemy's party member before fleeing battle. But maybe that could be a smaller pool of blood too. Maybe I'm overthinking it

I think if at least one enemy in the party is killed using C-4 and a blood crater is left behind just from that one prerequisite, that should be just fine I think.

TheMilkMan6669:
Additionally, I've also noticed that pits created by the Lode Gun are invisible when you create them, though they do still function as intended. Just not visible within the viewport.

PR #216, opened 2026-08-23
Moving SpeechSynthesizer Out of game.html, Adding Speech Modes

packardbell95:
As part of an effort to clean up `game.html`, the `SpeechSynthesizer` has been placed into its own file under `src/scripts/SpeechSynthesizer.js`

Additionally, the `SpeechSynthesizer` has been given modes which affect how the system handles new sentences:
- **Ignore:** New sentences will not be spoken if the voice is currently speaking
- **Interrupt:** New sentences will interrupt any current speech and be played instead
- **Queue:** All sentences will be spoken in order

The Game Settings menu has also been updated to accommodate these new modes. In place of the existing "Enable/Disable Speech" option, the option now allows players to cycle between Speech Synthesizer modes, including "Disabled." Options are persisted in `localStorage`

|  |
| --- |
| Cycling through Speech Synthesizer options in the Game Settings menu |


VocaPepper (commented):
Good, less mess. The Gex offender approves.

TheMilkMan6669 (approved):
Good call moving this stuff into its own separate script. Works as intended, the new settings are nice to have, we can merge.

PR #217, opened 2026-08-24
Update client API integrations for v4 server

VocaPepper:
This pull request introduces several major improvements to the API integration and authentication system. It updates all API endpoints to use new base URLs, adds a comprehensive account authentication system (including login, registration, recovery, and session management), and improves session state handling and error detection.

**API Endpoint Updates**
- All API calls in `tardAPI.js` and `pigeon.js` now use the new gateway URL and updated endpoint paths (e.g., `/start`, `/leaderboard`, `/status`), removing the `/api/` prefix.

**Authentication & Account Management (future use)**
- Introduced a full account system with methods for login, registration, logout, session checking, password recovery, and recovery code rotation. Integrates with your TardQuest Online account and supports rolling leaderboard updates.

**Miscellaneous Enhancements**
- Refactored PoW challenge storage into a dedicated helper function.

**API Server Updates**
- For information on the v4 server, please see [the documentation.](https://github.com/VocaPepper/tardquest-api/blob/4.0/docs/public/README.md)
- New Endpoint: `gateway.tardquest.online` replaces `vocapepper.com:9601`
- Old endpoint URLs automatically redirect to new server
- Client versions as low as 1.19.2 include limited API support. Leaderboard submissions will not be accepted on clients older than 4.0.2606

### DIFF
--- a/src/scripts/tardAPI.js
+++ b/src/scripts/tardAPI.js
@@ -23,10 +23,10 @@ const TardAPI = (function() {
     // --- Configuration ---
 
     /** @const {string} Base URL for all API endpoints */
-    const API_BASE = 'https://vocapepper.com:9601';
+    const API_BASE = 'https://gateway.tardquest.online';
 
     /** @const {string} Client API version (major.minor must match server) */
-    const CLIENT_API_VERSION = '3.2.260315';
+    const CLIENT_API_VERSION = '4.0.2606';
 
     /** @const {string} LocalStorage key for session ID persistence */
     const LS_SESSION_KEY = 'tardquestSID';
@@ -36,6 +36,9 @@ const TardAPI = (function() {
 
     /** @const {string} LocalStorage key for API feature toggle */
     const LS_API_FEATURES_KEY = 'tardquestApiFeaturesEnabled';
+
+    /** @const {string} LocalStorage key for authenticated account username */
+    const LS_ACCOUNT_KEY = 'tardquestAccountUser';
 
     // --- Logging Utility ---
 
@@ -50,6 +53,11 @@ const TardAPI = (function() {
 
     /** @type {string|null} Current session ID */
     let sessionId = sessionStorage.getItem(LS_SESSION_KEY) || null;
+
+    /** @type {string|null} Authenticated account username (null for guest sessions) */
+    let authUsername = (function() {
+        try { return localStorage.getItem(LS_ACCOUNT_KEY) || null; } catch { return null; }
+    })();
 
     /** @type {Object|null} Current PoW challenge data */
     let challengeData = null;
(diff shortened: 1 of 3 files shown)


src/scripts/tardAPI.js
```diff
@@ -51,6 +54,11 @@ const TardAPI = (function() {
     /** @type {string|null} Current session ID */
     let sessionId = sessionStorage.getItem(LS_SESSION_KEY) || null;
 
+    /** @type {string|null} Authenticated account username (null for guest sessions) */
+    let authUsername = (function() {
+        try { return localStorage.getItem(LS_ACCOUNT_KEY) || null; } catch { return null; }
```
packardbell95, line 59: **Suggestion, non-blocking:** I noticed that there are a lot of `localStorage` calls around API updates. All of them individually are wrapped in `try/catch` blocks. It might be worth creating a helper for these that handles the `try/catch` on its own

src/scripts/tardAPI.js
```diff
@@ -362,6 +350,19 @@ const TardAPI = (function() {
         }
     }
 
+    /**
+     * Validates current session with the API
+    /**
```
packardbell95, line 355: **Nitpick, non-blocking:** It almost looks like a nested comment block is starting here with the second `/**`

src/scripts/tardAPI.js
```diff
@@ -377,16 +378,21 @@ const TardAPI = (function() {
...
+                body: JSON.stringify({ session_id: sessionId, floor, level, exp: totalExp })
             });
 
             if (!res.ok) {
                 const data = await res.json().catch(() => ({}));
-                return { success: false, error: data.error || `HTTP ${res.status}` };
+                const errMsg = data.error || `HTTP ${res.status}`;
+                if (isSessionInvalidError(errMsg)) {
+                    log.warn('Session invalid/expired, clearing for re-creation');
+                    clearSession();
```
packardbell95, line 393: **Question:** Since this is a function that performs a validation check, should it be clearing the session itself? Shouldn't the caller of this function perform session clearing instead?

src/scripts/tardAPI.js
```diff
@@ -554,8 +562,8 @@ const TardAPI = (function() {
             if (limit) params.append('limit', String(limit));
 
             const url = params.toString()
-                ? `${API_BASE}/api/leaderboard?${params.toString()}`
-                : `${API_BASE}/api/leaderboard`;
+                ? `${API_BASE}/leaderboard?${params.toString()}`
```
packardbell95, line 565: **Comment, non-blocking:** This isn't a change related to this update, but since the line changed, I did notice that `params.toString()` is being called twice: once to check if it should be used and again to actually use it. It might be worth storing this in a `const` instead in a future update

packardbell95 (approved):
I left some feedback, none of which is critical. I haven't been able to test out the changes locally yet, but the changes overall appear to be pretty straightforward and look good to me 👍

TheMilkMan6669 (approved):
Things seem to be working functionally on my end. @VocaPepper I'll leave it up to you if you wanna merge now, or if you wanna touch up on the little things mentioned by @packardbell95 first.

